PR #209, opened 2014-01-26
[POC] Metadata and Annotation support

dantleech:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no (well, yes, but should be BC when finished) |
| Deprecations? | yes |
| Tests pass? | ? |
| Fixed tickets |  |
| License | MIT |
| Doc PR | not yet |

This PR adds initial support for Annotations:

``` php
use Symfony\Cmf\Bundle\RoutingBundle\Metadata\Annotations as CmfRouting;

/**
 * @PHPCRODM\Document(referenceable=true)
 *
 * @CmfRouting\Template("TestBundle:Content:index.html.twig")
 * @CmfRouting\Controller("cmf_content.controller:indexAction")
 */
class AnnotatedContent
```

For this I have introduced the `jms/metadata` package.

To maintain BC we would need to implement a DIC driver for the metadata factory.

Things to do:
- Add an XML driver (and I guess a YAML driver)
- Add the DIC driver to maintain BC

### DIFF
--- a/DependencyInjection/CmfRoutingExtension.php
+++ b/DependencyInjection/CmfRoutingExtension.php
@@ -39,6 +39,7 @@ class CmfRoutingExtension extends Extension
         if ($config['dynamic']['enabled']) {
             // load this even if no explicit enabled value but some configuration
             $this->setupDynamicRouter($config['dynamic'], $container, $loader);
+            $this->setupMetadataConfigurationDriver($config['dynamic'], $container);
         }
 
         /* set up the chain router */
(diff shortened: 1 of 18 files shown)


dbu:
ping @dantleech

dantleech:
Would be great, but don't know if I will have time in the near future..

dbu:
should we drop this one?

dantleech:
I think it would be a great feature. But we should try and do it for all bundles in the same way at the same time. But its still not something very high on my list, I don't see any harm in leaving it open as a reference / reminder?

dbu:
okay, no harm in keeping it open if its still relevant.

PR #240, opened 2014-04-20
[WIP] Add Twig functions for cmf_document_path and cmf_document_url

benglass:
Discussion in #239


dbu:
looks like this was the starting point for the whole Resource / ResourceBundle... but what do we do about those two twig functions? i still think we should have them.

PR #281, opened 2014-12-31
Cache the deps instead of the vendor directory

WouterJ:
Due to a bug in Composer, caching the vendor directory won't work very nice. (if `master` is aliased as `1.3-dev`, it'll cache the branch name instead of the alias, so it'll install the `master` branch, even if the alias changed to `2.0-dev`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ sudo: false
 
 cache:
     directories:
-        - vendor
+        - $HOME/.composer/cache
 
 php:
     - 5.3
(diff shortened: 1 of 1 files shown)


dbu:
somehow we try to drop a sqlite database triggering an error. not sure why exactly this is triggered with this PR

WouterJ:
@dbu the database is dropped by the Testing component: https://github.com/symfony-cmf/Testing/blob/master/src/Phpunit/DatabaseTestListener.php#L203-223

This happends for all tests since the first stable Testing release. What wonders me is why this is failing now and why it didn't before. Also, since when does sqlite not support dropping databases? :)

WouterJ:
Hmm, when rerunning the tests, they are even more broken than before. Shouldn't `--drop` only drop if the table did not yet exists? Also, reading the first and second error, it seems like something got really wrong:

````                                                                
  [Doctrine\DBAL\Exception\TableNotFoundException]                      
  An exception occurred while executing 'DROP TABLE phpcr_namespaces':  
  SQLSTATE[HY000]: General error: 1 no such table: phpcr_namespaces     
                                                                        
                                                                     
  [Doctrine\DBAL\Driver\PDOException]                                
  SQLSTATE[HY000]: General error: 1 no such table: phpcr_namespaces  
````

First, it tries to drop be can't since it did not yet exists, then it errors because the table did not exists...

WouterJ:
@dbu looks like this one is fixed by the jackalope pR

dbu:
yey! thanks!

PR #288, opened 2015-01-29
CMF Resource provider

dantleech:
This PR introduces a route provider for Puli resources via. the CMF
ResourceBundle.

Adding resource support means

- Possiblity to have context-aware route repositories, i.e. resolve different routes dending on the host or other critriea.
- Possiblity for cascading route resolution
- Possiblity for combining different route sources in the same route tree.


DependencyInjection/CmfRoutingExtension.php
```diff
@@ -289,6 +294,21 @@ public function loadSonataPhpcrAdmin($config, XmlFileLoader $loader, ContainerBu
...
+    {
+        $container->setParameter($this->getAlias() . '.resource.repository_name', $config['repository_name']);
+
+        if ($config['map_phpcr']) {
+            $container->setParameter($this->getAlias() . '.backend_type_phpcr', true);
+        }
+
+        if ($config['map_orm']) {
+            $container->setParameter($this->getAlias() . '.backend_type_orm', true);
+        }
```
dantleech: We have a small problem here: The resource bundle could potentially use either or both of the PHPCR/ORM Route document/entities. So I have added options to trigger the adding of those folders to doctrine.

However, we also need to have the "prefix" stuff working at least for the PHPCR-ODM documents.

Any ideas?

Resource/ResourceProvider.php
```diff
@@ -0,0 +1,134 @@
...
+    /**
+     * Find routes for the given glob path and
+     * add to the RouteCollection
+     *
+     * @param string          $path
+     * @param RouteCollection $collection
+     */
+    private function findRoutes($path, $collection)
+    {
+        $resources = $this->resourceRepository->find($path);
```
dantleech: So this works by first generating a list of glob paths, e.g.

For `/foo/bar/baz`

````php
array(
  '/',
  '/*',
  '/foo/*',
  '/foo/bar/*'
);
````

Is this about right?

Not very efficient, but could make this into a strategy later.

File: Resources/config/provider-resource.xml [hunk 3]
```diff
@@ -0,0 +1,22 @@
...
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <parameters>
+        <parameter key="cmf_routing.resource.provider.class">Symfony\Cmf\Bundle\RoutingBundle\Resource\ResourceProvider</parameter>
+    </parameters>
+
+    <services>
+
+        <service id="cmf_routing.route_provider" class="%cmf_routing.resource.provider.class%">
```
dantleech: We should probably switch to using aliases rather than each provider taking the main `cmf_routing.route_provider` ID.

dantleech:
This almost works, see the question above about prefixes.

dbu:
i have problems seeing how this is different from the phpcr router. i thought you check puli for a file and provide a route to download that file. but this is doing something else. i don't know enough of puli to exactly understand what the point of this is. can you explain that a bit please?

does this belong into this repo or into one of the resource repositories?

Resources/config/provider-resource.xml
dbu: agreed. want to do that in a separate PR to keep it apart? that would also allow to mix orm and phpcr routing - not necessarily a good idea but there sure is one usecase for that somewhere.

it will be a micro-bc break if somebody has a compiler pass to change the route provider service, as he will then hit the alias instead of the service definition, but depending on the service will still look the same.

Resource/ResourceProvider.php
dbu: hm, phpcr does this without the glob, so `/foo/bar/baz`, `/foo/bar`, `/foo`, `/` but you might even want to just support /foo/bar/baz directly. depends if you expect your routes to have a pattern that matches things in the url or just reflect the path

dantleech:
The way this is configured at the moment, it is functionaly equivilent to the PHPCR-ODM router as we have configured the `PhpcrRepository` directly.

However, it becomes more interesting when you use a more fancy repository, such as a "context aware" repository which would change the repositories base path depending on the host (multi-portal/site).

Or if you use a composite repository to combine several different route sources into a single tree.

Or you could use a cascading repository try a sequence of repositories, for example first try Site A, then Site B then fallback to Default Site.

Resource/ResourceProvider.php
dantleech: But I guess PHPCR-ODM adds a descendant selector?

Using the literal URLs without the wildcard would be more efficient, but at the cost of pattern matching - which people rarely need (right?). Good candidate for a strategy.

DependencyInjection/Configuration.php
```diff
@@ -118,6 +118,21 @@ public function getConfigTreeBuilder()
                                         ->scalarNode('manager_name')->defaultNull()->end()
                                     ->end()
                                 ->end()
+                                ->arrayNode('resource')
+                                    ->addDefaultsIfNotSet()
+                                    ->canBeEnabled()
+                                    ->children()
+                                        ->scalarNode('repository_name')->isRequired()->end()
+                                        ->booleanNode('map_orm')
+                                            ->info('Enable the RoutingBundle\'s Doctrine PHPCR Route Documents to be mapped')
```
lsmith77: I think you accidentally swapped the info text

lsmith77:
@webmozart BTW what is the ETA for a stable Puli release?

webmozart:
@lsmith77 February-March 2015. I want to publish a RC in February.

dbu:
@dantleech now that puli is released, any chance that you wrap this up?

dbu:
and does this still belong here or rather into the ResourceBundle?

dantleech:
I think this would depend on a release of the Resource lib. and bundle
so I think effort should go there first.

I am currently dedicating nearly all of my resources to another project
so don't have time to get back into Resource stuff atm, but maybe late
next month.

With respect to where this belongs, it probably doesn't belong here, at
least while the Resource stuff is still experimental. Is it in the scope
fo this bundle to provide third-party route providers?

dbu:
actually i feel the ResourceBundle would be the more natural fit for this. its all about resources. but of course in most real-world scenarios, one would want to use this provider with the CmfRoutingBundle to hook the provider into the dynamic router for example.

dbu:
i removed the milestone. if you agree, move the code to ResourceBundle and close this.

dantleech:
As I am moving the code now I see why I put it here in the first place.
The RoutingBundle already natively supports both the Doctrine ORM and
PHPCR-ODM route providers. Why shouldn't it, in theory, also support the
CMF/Puli resources?

WouterJ:
I actually believe in the future, the resource provider might take over both other route providers. Resource can handle PHPCR, ORM, filesystem, etc. natively.

dbu:
alright, then lets add it here. can you rebase and update the PR?

dantleech:
Rebased -- but merging would depend on a stable release of the Resource stuff and also I or somebody would have to invest time in this as it is probably non trivial.

WouterJ:
Hmm, I'm not sure if I really like the way this PR works. Basically, this PR lets you configure a repository for the routing stuff. This means we are limited to one repository and the RoutingBundle config is aware of the backend type (as that's the repository).

I'm more thinking about only 1 repository per application: The default repository. This can then be configured by the CmfResourceBundle. This allows you to for instance, bind multiple different backends to a default Composite repository.

Using the [Puli Discovery component](http://docs.puli.io/en/latest/discovery/introduction.html), the RoutingBundle can then find all routes:
```php
class ResourcesRouteProvider // ...
{
    // ...

    public function getRouteCollectionForRequest(Request $request)
    {
        $path = $request->getPathInfo();

        foreach ($this->getPaths($path) as $targetPath) {
            $this->findRoutes($targetPath, $collection);
        }
    }

    private function initializeRoutes($path, $collection)
    {
        $bindings = $this->discovery->findByPathAndType($path, 'cmf/routes');

        foreach ($bindings as $binding) {
            foreach ($binding->getResources() as $resource) {
                // ...
            }
        }
    }
}
```

Configuration then looks like:
```php
cmf_resource:
    repositories:
        routing_phpcr:
            type: doctrine_phpcr_odm
            basepath: /cmf/routing

        default:
            type: composite
            mounts:
                - { repository: routing_phpcr, mountpoint: /phpcr/routes }
    bindings:
        - { path: /phpcr/routes, type: cmf/routes }

cmf_routing:
    persistence:
        resource: ~
```

dantleech:
Could we not just have it default to `default` and keep the option to use an alternative repository?

WouterJ:
Yeah, we should still allow other repository names as well. However, I think tagging is really the implementation we need here, as it allows to have routes in multiple locations. E.g. simple cms documents and routes.

dantleech:
Not sure what you mean by "tagging"?

WouterJ:
Sorry, type binding. As shown in my comment above, using Puli Discovery.

dantleech:
ok yeah, so you would configure whatever repositories in the ResourceBundle, "bind" them to a Puli type and then let the Discovery component handle the rest in the RoutingProvider, sounds like the best way to do it.

webmozart:
In case this is relevant here: the Discovery component will be extended in beta8. Binding types will become actual PHP interface (or class) names such as `Cmf\Resource\ResourceBundle`. You'll be able to annotate this interface:

```php
/**
 * @BindingType
 * @Subject("RESOURCE")
 */
interface ResourceBundle {}
```

or continue to mark it as binding type with the CLI:

```
$ puli type --register Cmf\\Resource\\ResourceBundle
```

Furthermore and most important, it will be possible to bind class names:

```php
/**
 * @BindingType
 * @Subject("CLASS")
 */
interface Service {}
```

```
$ puli bind My\\ConcreteService Cmf\\Api\\Service
```

This could be useful here.

WouterJ:
Especially that second feature would be usefull :) (I assume the code of `puli bind` will be available as a normal PHP method that we can use in the ResourceBundle?)

webmozart:
@WouterJ of course, same as now :)

webmozart:
See https://github.com/puli/issues/issues/114 for more detail

dbu:
@WouterJ @dantleech is this a blocker for the 2.0 bundle? i guess we can't really do it in a BC way so if its not in 2.0 it will have to wait for quite a while. but i really want to get 2.0 moving to get symfony 3 support.

dantleech:
@dbu I see this as a feature, not a BC break. We would just be adding a new Provider .. ? In which case we can release 2.0 and put this in 2.1 if it turns out to be a worthy addition.

WouterJ:
This should be closed in favor of https://github.com/symfony-cmf/routing-bundle/pull/314 btw

dantleech:
oh, didn't see that one :) well done.

PR #292, opened 2015-03-14
Migrate errorElement

EmmanuelVella:
The last one ;)


dbu:
to merge this, we would need to update the composer.json to conflict with too old sonata admin versions.

dbu:
and then i have the same question as for blockbundle: when will sonata admin 2.4 be released?

EmmanuelVella:
Hum this means adding something like ``"sonata-project/core-bundle": "~2.3,>=2.3.1"`` in the ``require`` section ?

dbu:
not exactly. there is a specific [conflicts](https://getcomposer.org/doc/04-schema.md#conflict) section. there we would say 

    conflicts: {
         "sonata-project/core-bundle": "<2.3.1"
    },

EmmanuelVella:
Thank you very much David, I didn't know this section. That's great. I have updated my PR accordingly.

dbu:
@WouterJ what do you think, shall we merge this and aim the next release to target the new sonata admin already?

oh, and with the conflict thing in composer, i see one more reason for @dantleech proposal to move the admins to its own repository. if you don't use the sonata admin classes we provide here, but some custom admin built on an older version of sonata, there is no conflict. i created https://github.com/symfony-cmf/symfony-cmf/issues/216 to discuss this.

composer.json
```diff
@@ -27,6 +27,9 @@
         "symfony/monolog-bundle": "~2.3",
         "doctrine/orm": "~2.3"
     },
+    "conflict": {
+      "sonata-project/core-bundle": "<2.3.1"
```
WouterJ: missing 2 spaces in the indent (we use 4 spaces indentation)

WouterJ:
Just to be clear: This means we are no longer compatible with SonataAdmin 2.3 (as it's still using the [old namespace](https://github.com/sonata-project/SonataAdminBundle/blob/2.3/Admin/Admin.php#L32))? In that case, -1.

EmmanuelVella:
@WouterJ Thank you, I just fixed the indentation.

No, we are style compatible with SonataAdminBundle / SonataAdminBundle >=2.3.1, as the new namespace is available since this version. The old namespace was recently removed on the 2.4 branch (see https://github.com/sonata-project/SonataAdminBundle/commit/ccd2248ebeaf50e7d30df2ba37169488594ebd1d)

dbu:
the build on travis indicates an error with the dependencies.

EmmanuelVella:
I have updated composer.json to use ``sonata-project/doctrine-phpcr-admin-bundle: 2.0.*@dev``, tests are ok now !

WouterJ:
That's exactly what I meant :) We are only compatible with SonataAdminBundle ~2.4 here.

The problem is that the typehint of SonataAdminBundle 2.4 has changed to the new namespace (https://github.com/sonata-project/SonataAdminBundle/blob/master/Admin/Admin.php#L584). This means that a method overriding this function will be incompatible when it's not updating this typehint to the new class. At the same time, updating this typehint to the new class means the method is incompatible with its parent signature for SonataAdminBundle 2.3, as it's still using the old typehint.

So there is no way to make this work on 2.4 *and* <=2.3. I'm -1 for breaking 2.3 compatibility, as we cannot force people to update their sonata admin packages.

EmmanuelVella:
@WouterJ You are completely right. But that's so sad to not be able to use the bundle just because of this !

Maybe I should create a custom external validator so we don't have to use the ``Admin::validate()`` method anymore ?

EmmanuelVella:
I have updated my PR to replace the ``RouteAdmin::validate`` method with a custom ``RouteDefaultsValidator``.

This way the bundle is no longer dependant of the ``ErrorElement`` namespace, and is compatible with both ``2.3`` and ``2.4`` admin versions.

WouterJ:
:+1:

WouterJ:
refs #296

dbu:
excellent idea!

actually it would be cool to convert all validation to this approach and get rid of sonata specific validation alltogether. this way, validation is reusable regardless of the tools used.

EmmanuelVella:
Thanks !

Admin/RouteAdmin.php
```diff
@@ -117,11 +109,6 @@ public function setContentRoot($contentRoot)
         $this->contentRoot = $contentRoot;
     }
 
-    public function setControllerResolver($controllerResolver)
```
dbu: hm, this is a BC break for the SimpleCmsBundle: https://travis-ci.org/symfony-cmf/SimpleCmsBundle/jobs/48334145

what should we do? re-add an empty method?

EmmanuelVella: If you can't remove the setter injection (https://github.com/symfony-cmf/SimpleCmsBundle/blob/1.2.0/Resources/config/admin-phpcr.xml#L36) for BC reasons, I think that adding an empty method (maybe with a deprecated annotation) is the solution.

PR #293, opened 2015-03-16
ORM content repository implementation ready

wolfwolker:
|Q|A|
|---|---|
|Bug fix?|yes
|New feature?|yes
|BC breaks?|no
|Deprecations?|no
|Tests pass?|?
|Fixed tickets|	
|License|MIT
|Doc PR|not yet

ORM content repository implementation was already included in the bundle, 
but not well declared in provider-orm.xml,
nor loaded in the extension.

### DIFF
--- a/Resources/config/provider-orm.xml
+++ b/Resources/config/provider-orm.xml
@@ -12,7 +12,7 @@
 
     <services>
 
-        <service id="cmf_routing.content_repository" class="%cmf_routing.orm.content_repository.class%">
+        <service id="cmf_routing.orm_content_repository" class="%cmf_routing.orm.content_repository.class%">
             <argument type="service" id="doctrine"/>
             <call method="setManagerName"><argument>%cmf_routing.dynamic.persistence.orm.manager_name%</argument></call>
         </service>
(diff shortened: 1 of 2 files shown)


Resources/config/provider-orm.xml
```diff
@@ -12,7 +12,7 @@
 
     <services>
 
-        <service id="cmf_routing.content_repository" class="%cmf_routing.orm.content_repository.class%">
+        <service id="cmf_routing.orm_content_repository" class="%cmf_routing.orm.content_repository.class%">
```
dantleech: Should be `cmf_routing.orm.content_repository` I think.

wolfwolker: I followed the same name pattern than in provider-phpcr.xml (https://github.com/symfony-cmf/RoutingBundle/blob/master/Resources/config/provider-phpcr.xml#L33)

dbu:
indeed, you are right that the content repository loading was incomplete. thanks a lot for hunting this issue down @wolfwolker ! does this fully work for you like this? if you have updates to the documentation, we would appreciate it - the phpcr-odm version is much more tested than the orm, but the intention is to support both.

i will merge as is. the service naming is consistent with the rest. the naming of both phpcr and orm provider services is not nice, but changing them is a BC break we can only do with version 2.0.

wolfwolker:
Our intention is to include some cmf parts in our orm projects. Firstly
routing & autorouting. So we are going to test them and add orm support
where its incomplete. So in near future will send more contributions.

Cheers.
Ps:  Thanks for your work. The whole Cmf project is amazing.

dbu:
glad if the cmf is useful. looking forward to further cleanups on features or documentation.

PR #295, opened 2015-03-21
Allow specification of custom URL generator

dantleech:
This PR allows the specification of a custom URL generator.

````
cmf_routing:
    dynamic:
        url_generator: my_service_id
````


dantleech:
The use case here is that we are not currently using the RouteReferrer interface and also we are caching URLs in the content nodes, so I think using a custom URL generator is the way forward.

Another option would be to add a new Router which just implements the URL generation.

dbu:
in the interest of modularity, i think this is the correct thing to do. thanks!

PR #297, opened 2015-04-18
revert BC break introduced in #292

dbu:
removing the setControllerResolver method leads to problems when a service definition tries to call this method.
/cc @EmmanuelVella

### DIFF
--- a/Admin/RouteAdmin.php
+++ b/Admin/RouteAdmin.php
@@ -15,6 +15,7 @@ use Sonata\AdminBundle\Datagrid\DatagridMapper;
 use Sonata\AdminBundle\Datagrid\ListMapper;
 use Sonata\AdminBundle\Form\FormMapper;
 use Sonata\DoctrinePHPCRAdminBundle\Admin\Admin;
+use Symfony\Component\HttpKernel\Controller\ControllerResolverInterface;
 use Symfony\Cmf\Bundle\RoutingBundle\Model\Route;
 use PHPCR\Util\PathHelper;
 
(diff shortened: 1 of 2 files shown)


Resources/config/admin-phpcr.xml
```diff
@@ -32,6 +32,10 @@
             <call method="setRouteRoot">
                 <argument>%cmf_routing.dynamic.persistence.phpcr.admin_basepath%</argument>
             </call>
+
+            <call method="setControllerResolver">
+                <argument type="service" id="controller_resolver" />
+            </call>
```
dbu: i know this is unused in the base class, but as it was protected, extending classes might rely on it.

Admin/RouteAdmin.php
```diff
@@ -109,6 +117,14 @@ public function setContentRoot($contentRoot)
         $this->contentRoot = $contentRoot;
     }
 
+    /**
+     * @deprecated The resolver should not be used directly, but the RouteDefaults validation.
```
WouterJ: I think we should always show the version in which it was deprecated

dbu:
good point about noting the version in deprecated annotation. updated accordingly.

Admin/RouteAdmin.php
```diff
@@ -109,6 +117,14 @@ public function setContentRoot($contentRoot)
         $this->contentRoot = $contentRoot;
     }
 
+    /**
+     * @deprecated Since 1.4, the resolver should not be used directly, but the RouteDefaults validation.
```
WouterJ: this doesn't sound like a complete sentence: "the resolver should not be used directly, but te RouteDefaults validation (...?)"

dbu: better?

WouterJ:
:+1: :)

EmmanuelVella:
:+1: Thank you for fixing this.

PR #301, opened 2015-05-23
Generate candidates from decoded URLs

eggyal:
Symfony\Component\Routing\Matcher\UrlMatcher::match() performs a rawurldecode() on a request's pathInfo in order to find matching routes, so one must do the same when generating candidate routes.

### DIFF
--- a/Doctrine/Phpcr/PrefixCandidates.php
+++ b/Doctrine/Phpcr/PrefixCandidates.php
@@ -103,7 +103,7 @@ class PrefixCandidates extends Candidates
     public function getCandidates(Request $request)
     {
         $candidates = array();
-        $url = $request->getPathInfo();
+        $url = rawurldecode($request->getPathInfo());
         foreach ($this->getPrefixes() as $prefix) {
             $candidates = array_unique(array_merge($candidates, $this->getCandidatesFor($url, $prefix)));
         }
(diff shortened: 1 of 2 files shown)


lsmith77:
hmm .. can you add a test case to illustrate the issue

lsmith77:
thx .. the test failures are not related to these changes

eggyal:
@lsmith77: Is that just a comment on the Travis failures, or are you saying that my test case is unrelated to my fix?

lsmith77:
build is fixed

lsmith77:
@dbu fine for you as well?

dbu:
ah, makes sense indeed. this makes me wonder: what about spaces in urls when we generate? if we match urls with spaces but generation relies on the browser to url encode spaces, this would be awkward. i hope i find some time to play around in the cmf sandbox today to see what happens. also with core symfony routes in the configuration...

eggyal:
@dbu: The UrlGenerator percent-encodes: see https://github.com/symfony/Routing/blob/master/Generator/UrlGenerator.php#L192.  I suppose there is an argument that this ought to be optional (e.g. if people actually *want* to serve IRIs), but then there's always the possibility of extending/overriding that behaviour.

dbu:
thanks a lot!

eggyal:
@lsmith77 @dbu : Are you able to give any idea when this might make it into a release?

dbu:
sorry, we really lag behind with releasing. we hope to wrap things up early september. the milestone for the release looks like this: https://github.com/symfony-cmf/RoutingBundle/milestones/1.4 but we might just throw out things if we do not manage to finish them

PR #303, opened 2015-06-23
fix bug in sonata admin

dbu:
when locale pattern is activated, add a field for the locale, not for the format


lsmith77:
BTW this also affected earlier versions but I guess its enough to fix it in 1.3

lsmith77:
https://github.com/symfony-cmf/RoutingBundle/releases/tag/1.3.5

dbu:
> BTW this also affected earlier versions but I guess its enough to fix it
> in 1.3

yes, but i thought we don't maintain older releases. its not a security
issue, just a nuisance so if people want it fixed, they should update imho.

we will need to merge the 1.3 branch to master at some point however, to
get the fix there.

PR #304, opened 2015-06-25
fix issues with PHP7 due to use of String annotation class

lsmith77:


composer.json
```diff
@@ -21,6 +21,7 @@
     "require-dev": {
         "symfony-cmf/core-bundle": "~1.1",
         "symfony-cmf/testing": "~1.2,>=1.2.6",
+        "doctrine/phpcr-odm": "~1.3@dev",
```
WouterJ: there is no need for the additional `@dev` flag

PR #306, opened 2015-08-23
persisting route condition as well

### LINKED ISSUE #270: also persist "condition" Route property
follow up for https://github.com/symfony-cmf/RoutingBundle/issues/267

dbu:
fix #270

adding a metadata listener to keep symfony 2.3 support for now. this listener can be removed again when we drop symfony 2.7 support.

### DIFF
--- a/Doctrine/RouteConditionMetadataListener.php
+++ b/Doctrine/RouteConditionMetadataListener.php
@@ -1,0 +1,74 @@
+<?php
+
+/*
+ * This file is part of the Symfony CMF package.
+ *
+ * (c) 2011-2015 Symfony CMF
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Cmf\Bundle\RoutingBundle\Doctrine;
+
+use Doctrine\Common\EventSubscriber;
+use Doctrine\Common\Persistence\Event\LoadClassMetadataEventArgs;
+use Doctrine\ORM\Mapping\ClassMetadata as OrmClassMetadata;
+use Doctrine\ODM\PHPCR\Mapping\ClassMetadata as PhpcrClassMetadata;
+
+/**
+ * Metadata listener to remove mapping for condition field if the field does not exist.
+ *
+ * The condition option was only added in Symfony 2.4 and is missing from 2.3.
+ * When we drop Symfony 2.3 support, this listener can be dropped.
+ *
+ * @author David Buchmann <mail@davidbu.ch>
+ */
+class RouteConditionMetadataListener implements EventSubscriber
+{
+    /**
+     * @return array
+     */
+    public function getSubscribedEvents()
+    {
+        return array(
+            'loadClassMetadata',
+        );
+    }
+
+    /**
+     * Handle the load class metadata event: remove translated attribute from
+     * fields and remove the locale mapping if present.
+     *
+     * @param LoadClassMetadataEventArgs $eventArgs
+     */
+    public function loadClassMetadata(LoadClassMetadataEventArgs $eventArgs)
+    {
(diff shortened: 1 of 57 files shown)


Doctrine/RouteConditionMetadataListener.php
```diff
@@ -0,0 +1,72 @@
...
+
+        if ($meta instanceof OrmClassMetadata) {
+            /** @var $meta OrmClassMetadata */
+            unset($meta->fieldMappings['condition']);
+        } elseif ($meta instanceof PhpcrClassMetadata) {
+            /** @var $meta PhpcrClassMetadata */
+            unset($meta->mappings['condition']);
+            foreach ($meta->fieldMappings as $index => $name) {
+                if ('condition' === $name) {
+                    unset($meta->fieldMappings[$index]);
```
dbu: it seems for phpcr-odm 2.0 we could try to be more consistent with the orm. this is less elegant...

Doctrine/RouteConditionMetadataListener.php
```diff
@@ -0,0 +1,72 @@
...
+
+        $meta = $eventArgs->getClassMetadata();
+        if (!$meta->getReflectionClass()->isSubclassOf('Symfony\Component\Routing\Route')) {
+            return;
+        }
+
+        if ($meta instanceof OrmClassMetadata) {
+            /** @var $meta OrmClassMetadata */
+            unset($meta->fieldMappings['condition']);
+        } elseif ($meta instanceof PhpcrClassMetadata) {
```
dbu: yes, we could split this in two classes instead of the instanceof check. probably with a common base class as most is the same. but i think its just not worth it, the whole listener can be removed again when we drop symfony 2.3 support.

Doctrine/RouteConditionMetadataListener.php
```diff
@@ -0,0 +1,72 @@
+<?php
+
+/*
+ * This file is part of the Symfony CMF package.
+ *
+ * (c) 2011-2014 Symfony CMF
```
WouterJ: should be 2015

WouterJ:
The listener doesn't seem to work (both Sf 2.3 tests fail because of missing property).

Doctrine/RouteConditionMetadataListener.php
dbu: this is all over the place. updated for this bundle now.

Doctrine/RouteConditionMetadataListener.php
```diff
@@ -0,0 +1,72 @@
...
+    {
+        if (property_exists('Symfony\Component\Routing\Route', 'condition')) {
+            return; // nothing to do
+        }
+
+        $meta = $eventArgs->getClassMetadata();
+        if (!$meta->getReflectionClass()->isSubclassOf('Symfony\Component\Routing\Route')) {
+            return;
+        }
+echo 'unset condition from metadata';
```
dbu: lets see if there is a problem with how i unset the mapping or with the subscriber

dbu:
i wont have time to wrap this up before my holidays. if somebody can pick this up i am glad, otherwise i will try to finish this in 3 weeks.

WouterJ:
@dbu the problem here is that during the loading of the metadata by the driver, the metadata is already validated. This results in the exception. The loadClassMetadata event is dispatched after the driver loaded the metadata

dbu:
what can we do? invert the logic and not write the mapping in the config file but *add* the mapping for condition if we are on a new enough version of symfony?

WouterJ:
If that works, it sounds like a plan

dbu:
alright, i added a test and things look like they can work this way. lets see if travis agrees for all combinations of versions.

WouterJ:
:+1:

PR #308, opened 2015-09-30
Fix problem with reserved keywords in mysql

edgartebar:


dbu:
ups! thanks for this fix. do i get it correctly that this is no BC breaks for postgres / sqlite users, as the column will still be called the same as before?

edgartebar:
Doctrine suports quote strategy, i supose if doctrine suports postgres and sqlite it will work. http://doctrine-orm.readthedocs.org/en/latest/reference/basic-mapping.html#quoting-reserved-words

lsmith77:
hmm not sure if this isn't going to bite us ... keyword quoting is quite painful. maybe better to just rename ..

almacbe:
@lsmith77 @dbu opt you, but nowadays does not work for ORM :-P please tell us what you do prefer and we change the PR please. Thanks

gerfigna:
+1

dbu:
the condition mapping was only added a week ago or so, and no version tagged with it. lets change it to `condition_expr` i would say, to make peoples live easier. does that make sense @lsmith77 ?

edgartebar:
Done, i sure that i will live easier XD thanks

lsmith77:
+1

dbu:
thanks a lot!

PR #311, opened 2015-10-28
Fix route basepaths configuration

WouterJ:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | https://github.com/symfony-cmf/RoutingBundle/issues/290
| License       | MIT
| Doc PR        | https://github.com/symfony-cmf/symfony-cmf-docs/pull/715

* Deprecated the `route_basepath` setting in favor of `route_basepaths`
* Allowed string values for `route_basepaths`, in case only one basepath needs to be configured
* Configuring `route_basepaths` overrides the default `/cms/routes` value, add it explicitely to include it in the basepath list
* Multiple `route_basepaths` configuration in multiple config files results in merging all configured basepaths.
* Make sure no duplicated basepaths are configured
* Improve testing

/cc @dbu

### DIFF
--- a/DependencyInjection/CmfRoutingExtension.php
+++ b/DependencyInjection/CmfRoutingExtension.php
@@ -240,13 +240,15 @@ class CmfRoutingExtension extends Extension
 
         $container->setParameter(
             $this->getAlias() . '.dynamic.persistence.phpcr.route_basepaths',
-            $config['route_basepaths']
+            array_values(array_unique($config['route_basepaths']))
         );
 
-        $basePath = reset($config['route_basepaths']);
+        /**
+         * @deprecated The cmf_routing.dynamic.persistence.phpcr.route_basepath parameter is deprecated as of version 1.4 and will be removed in 2.0. Use the cmf_routing.dynamic.persistence.phpcr.route_basepaths parameter instead.
+         */
         $container->setParameter(
             $this->getAlias() . '.dynamic.persistence.phpcr.route_basepath',
-            $basePath
+            reset($config['route_basepaths'])
         );
 
         $container->setParameter(
@@ -288,7 +290,7 @@ class CmfRoutingExtension extends Extension
             return;
         }
 
-        $basePath = empty($config['admin_basepath']) ? reset($config['route_basepaths']) : $config['admin_basepath'];
+        $basePath = $config['admin_basepath'] ?: reset($config['route_basepaths']);
         $container->setParameter($this->getAlias() . '.dynamic.persistence.phpcr.admin_basepath', $basePath);
 
         $loader->load('admin-phpcr.xml');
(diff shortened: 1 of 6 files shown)


DependencyInjection/Configuration.php
```diff
@@ -81,13 +81,25 @@ public function getConfigTreeBuilder()
                                         ->ifTrue(function ($v) { isset($v['route_basepath']) && isset($v['route_basepaths']); })
                                         ->thenInvalid('Found values for both "route_basepath" and "route_basepaths", use "route_basepaths" instead.')
                                     ->end()
+                                    ->beforeNormalization()
+                                        ->ifTrue(function ($v) { return isset($v['route_basepath']); })
+                                        ->then(function ($v) {
+                                            @trigger_error('The route_basepath setting is deprecated as of version 1.4 and will be removed in 2.0. Use route_basepaths instead.', E_USER_DEPRECATED);
```
WouterJ: The deprecation handling system in Symfony is now good enough to immediately trigger deprecation notices. As the `@` is used, these errors will be silent by default (unless someone uses a custom error handler, which is only the case when you use the PHPunit Bridge or the deprecation catcher from the web dev toolbar, in both cases one is looking for reports about deprecated feature usages).

DependencyInjection/CmfRoutingExtension.php
```diff
@@ -240,13 +240,15 @@ public function loadPhpcrProvider($config, XmlFileLoader $loader, ContainerBuild
 
         $container->setParameter(
             $this->getAlias() . '.dynamic.persistence.phpcr.route_basepaths',
-            $config['route_basepaths']
+            array_values(array_unique($config['route_basepaths']))
```
WouterJ: array_values call is needed because the array keys need to match as well, in order to let the tests pass.

DependencyInjection/Configuration.php
```diff
@@ -81,13 +81,25 @@ public function getConfigTreeBuilder()
...
+
+                                            return $v;
```
WouterJ: `route_basepath` -> `route_basepaths` transformation is already handled by the fixXmlConfig feature.

WouterJ:
Build failure is unrelated

DependencyInjection/CmfRoutingExtension.php
dbu: would this not be the job of the test? but well, should not hurt mutch to elminiate holes in the keys, could lead to confusion somewhere.

dbu:
oh, composer timing out in lowest...

great job! i added the pull request template in the top. i think the only missing thing is documentation. can you add a PR for the doc, then we can merge this.

WouterJ:
Done, see https://github.com/symfony-cmf/symfony-cmf-docs/pull/715

dbu:
thanks a lot!

PR #312, opened 2015-10-31
Clean up DI Extension & Configuration classes

WouterJ:
Reasoning
---

The DI extension was very big and hard to follow. It's still pretty big, but I tried to simplify it a bit by moving some validation to the Configuration and splitting the handling in some methods.

Changes included in this PR
---

 * Made all methods in the extension class private instead of public (except from `load()`). These methods shouldn't be overriden and I don't think this is a BC break as nobody will ever override a DI extension (or they must do something very hacky).
 * Changed all `$this->getAlias()` calls to just `"cmf_routing"`. I don't see any advantage of calling this method: The `cmf_routing` prefix is already hardcoded in the service configuration files and many other places. To me, it only adds many method calls and some confusion (especially when used inside a `setAlias()` call of a service definition).
 * Split building the config tree in multiple methods, making it easier to follow what's going on.
 * Fixed the deprecation trigger of `route_basepath` when using XML.

Info
---

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

### DIFF
--- a/DependencyInjection/CmfRoutingExtension.php
+++ b/DependencyInjection/CmfRoutingExtension.php
@@ -227,12 +227,12 @@ class CmfRoutingExtension extends Extension
 
     private function loadSonataPhpcrAdmin($config, XmlFileLoader $loader, ContainerBuilder $container)
     {
-        $loader->load('admin-phpcr.xml');
-
         $bundles = $container->getParameter('kernel.bundles');
         if ('auto' === $config['use_sonata_admin'] && !isset($bundles['SonataDoctrinePHPCRAdminBundle'])) {
             return;
         }
+
+        $loader->load('admin-phpcr.xml');
 
         $basePath = $config['admin_basepath'] ?: reset($config['route_basepaths']);
         $container->setParameter('cmf_routing.dynamic.persistence.phpcr.admin_basepath', $basePath);
--- a/DependencyInjection/CmfRoutingExtension.php
+++ b/DependencyInjection/CmfRoutingExtension.php
@@ -22,6 +22,7 @@ use Symfony\Component\Config\Loader\LoaderInterface;
 /**
  * @author Philippo de Santis
  * @author David Buchmann
+ * @author Wouter de Jong <wouter@wouterj.nl>
  */
 class CmfRoutingExtension extends Extension
 {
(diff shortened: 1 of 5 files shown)


DependencyInjection/CmfRoutingExtension.php
```diff
@@ -75,102 +79,81 @@ public function setupFormTypes(array $config, ContainerBuilder $container, Loade
...
-        $container->setParameter($this->getAlias() . '.uri_filter_regexp', $config['uri_filter_regexp']);
-        $container->setParameter($this->getAlias() . '.route_collection_limit', $config['route_collection_limit']);
-
-        $locales = empty($config['locales']) ? array() : $config['locales'];
-        $container->setParameter($this->getAlias() . '.dynamic.locales', $locales);
-        if (count($locales) === 0 && $config['auto_locale_pattern']) {
+        $container->setParameter('cmf_routing.default_controller', $defaultController);
+
+        $locales = $config['locales'];
+        if (0 === count($locales) && $config['auto_locale_pattern']) {
```
dantleech: Just reviewed some code and said that there should always be a blank line before IF statements, but this groups a concern, so I would say it was OK. But -- why do we need to assign to a locale variable anyway?

WouterJ: Yeah, I was not sure about it as well. We pass it to the `load*Persistence` methods below, but we can also just pass `$config['locales']` to these methods.

DependencyInjection/CmfRoutingExtension.php
```diff
@@ -232,49 +198,25 @@ private function setupDynamicRouter(array $config, ContainerBuilder $container,
...
-        $container->setAlias(
-            $this->getAlias() . '.content_repository',
-            $this->getAlias() . '.phpcr_content_repository'
-        );
-
-        if (!$locales) {
-            $container->removeDefinition($this->getAlias() . '.phpcrodm_route_locale_listener');
+        $container->setParameter('cmf_routing.dynamic.persistence.phpcr.route_basepath', reset($config['route_basepaths']));
+        $container->setParameter('cmf_routing.dynamic.persistence.phpcr.content_basepath', $config['content_basepath']);
+        $container->setParameter('cmf_routing.dynamic.persistence.phpcr.manager_name', $config['manager_name']);
```
dantleech: Why not use your `configureParameters` method here?

Resources/config/provider-phpcr.xml
```diff
@@ -22,6 +22,8 @@
             <call method="setRouteCollectionLimit"><argument>%cmf_routing.route_collection_limit%</argument></call>
         </service>
 
+        <service id="cmf_routing.route_provider" alias="cmf_routing.phpcr_route_provider"/>
```
dantleech: What is the reason for adding these service definitions?

WouterJ: They were first done in the extension, but it makes no sense to do it in the extension (as they are just static).

dantleech:
Big +1

dbu:
great work! thanks a lot wouter.it looks sane to me, but its a lot of change. did you try it out in the sandbox to see if there is any regression we can note there? once we did that, lets just merge and wait for users to report issues, if there are any glitches :-)

DependencyInjection/CmfRoutingExtension.php
dbu: ping @WouterJ

dbu:
would love to merge this one. did you test it in the cmf sandbox?

WouterJ:
I've tested this today with SE & Sandbox (very minor testing though, but this worked as expected after fixing a minor bug)

DependencyInjection/CmfRoutingExtension.php
dbu: to me that does not need to be a fix rule. the important is that things that logically belong together should visually stick together. and i would not do the assignment if we only use the value a few times as argument.

dbu:
lets ignore styleci in this branch. i guess we get the fixes in the dedicated styleci PR - from what it complains, its pretty massive...

tests look good, except lowest died out on travis.

PR #313, opened 2015-10-31
Trigger deprecation notices

WouterJ:
Symfony does now have really nice tools and a good handling of deprecation notices. So we should trigger deprecation notices whenever we deprecate anything.

Besides that, this 1.x release is probably one of the last releases in the 1 major so it also makes sense to warn users about deprecated features.

/cc @lsmith77 @dbu @dantleech

### DIFF
--- a/Doctrine/Phpcr/RedirectRoute.php
+++ b/Doctrine/Phpcr/RedirectRoute.php
@@ -73,9 +73,9 @@ class RedirectRoute extends RedirectRouteModel implements PrefixInterface, Hiera
      */
     public function setParent($parent)
     {
-        $this->parent = $parent;
+        @trigger_error('The Route#setParent() method is deprecated as of version 1.2 and will be removed in 2.0. Use setParentDocument() instead.', E_USER_DEPRECATED);
 
-        return $this;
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -37,7 +37,7 @@ before_install:
   - composer self-update
   - composer require symfony/symfony:${SYMFONY_VERSION} --no-update
   
-install: composer update --prefer-source $COMPOSER_FLAGS
+install: composer update -vvv --prefer-source $COMPOSER_FLAGS
```
WouterJ: I hope this gives us enough time, so we can avoid using `travis_wait`.

dbu: no luck. can we do something to reduce the options space? like not allow symfony > 2.3 in the lowest build?

WouterJ: We already do that. Seems like it's stuck resolving jackalope, maybe we can narrow that down as well

dbu:
still no luck. argh! can we do more? more specific versions without stars?

WouterJ:
Removed the build fixes from this PR, let's merge this and fix the lowest deps build in another PR.

dbu:
:+1:

PR #314, opened 2015-10-31
[2.0][POC] Start with a resource provider

WouterJ:
Just to play around with fully implementing Puli in CMF 2.0. In the future, I think we can remove all persistence implementations and just only rely on resource (people will configure resource how to handle resources, e.g. with a filesystem, PHPCR, ORM, etc.).

This requires https://github.com/symfony-cmf/ResourceBundle/pull/16

Replaces https://github.com/symfony-cmf/RoutingBundle/pull/288. This PR uses Puli discovery to locate resources, this deprecates all basepath stuff (:tada:).

Example Config
---
```
cmf_resource:
    repositories:
        # configure CMF Resource to use a PHPCR ODM backend
        default:
            type: doctrine_phpcr_odm
    discovery:
        bindings:
            # bind the /cmf/simple resource and its children to the
            # cmf/routes type (discovered by the ResourceRouteProvider)
            - { path: /cms/simple, type: cmf/routes }
            - { path: /cms/simple/*, type: cmf/routes }

cmf_routing:
    dynamic:
        persistence:
            # PHPCR is required for mapping, this has to be refactored when we start developing CMF 2
            phpcr: true
            resource: true
        templates_by_class:
            Symfony\Cmf\Bundle\SimpleCmsBundle\Doctrine\Phpcr\Page: AcmeDemoBundle:Page:index.html.twig
```


Provider/ResourceRouteProvider.php
```diff
@@ -0,0 +1,111 @@
...
+        return $this->candidatesStrategy->getCandidates($request);
+    }
+
+    private function discoverRoutes()
+    {
+        if (0 !== count($this->routesByUri)) {
+            return;
+        }
+
+        foreach ($this->discovery->findBindings('cmf/routes') as $binding) {
```
WouterJ: it's not so nice to do this every request, we should add some caching here.

dantleech:
This looks awesome :)

Provider/ResourceRouteProvider.php
dantleech: So we eagerly load all the routes in the resource tree into memory? What about depth > 1 routes? e.g. `/cms/simple/*/**`

dbu:
does this block the 2.0 release? can it be added in 2.1 or is it too much of a BC break?

WouterJ:
@dbu it's not a BC break at all, adding new providers isn't a BC break...

However, this feature requires Puli. All CMF 2.0 bundles require SonataDoctrinePhpcrAdminBundle 2.0, which requires Puli. So in some way, merging this feature in 2.0 or not, we all have to wait untill Puli to release CMF 2.0.

dbu:
> @dbu <https://github.com/dbu> it's not a BC break at all, adding new
> providers isn't a BC break...

ah, i thought it replaces the existing providers. great, then no blocker.

> All CMF 2.0 bundles require
> SonataDoctrinePhpcrAdminBundle 2.0, which requires Puli. So in some way,
> merging this feature in 2.0 or not, we all have to wait untill Puli to
> release CMF 2.0.

i hope we can solve this with the effort by max to move all admins to
the admin integration bundle.

WouterJ:
> i hope we can solve this with the effort by max to move all admins to the admin integration bundle.

Yeah, but that means releasing CMF 2.0 without any stable admin integration. That seems a bit strange to me.

dbu:
not everybody uses/needs sonata admin, i would think, so still better than not releasing a version that works with symfony 3.

dantleech:
we could also implement the resource repository / resource APIs natively in 2.0 until Puli stable, if we really had to - that would be enough for the tree browser / resource rest use case (although not for this PR  which requires discovery etc).

dbu:
what do we do with this? is it a blocker for 2.0 or not?

wouterj:
No, can be easily skipped and moved to 2.1.

PR #315, opened 2015-11-04
Fix lowest deps build

WouterJ:


dbu:
looks like we need a newer version of jackalope?

WouterJ:
@dbu the strange thing is that the versions are exactly the same as in https://travis-ci.org/symfony-cmf/RoutingBundle/jobs/89195242 and it works correctly there...

dbu:
@WouterJ i tried with jackalope 1.2. i think we should go with that version. now we end up with a missing orm command class. i tried upping orm to 2.5 in the require-dev and hope this will make that problem go away.

dbu:
hm. now docrine 2.5 requires php  5.4 at least. am i mistaken that we need orm 2.5? i would be ok to raise the php version to 5.4, 5.3 is horribly old.

PR #316, opened 2015-11-21
Improve testing speed by reusing the database

WouterJ:
I'm still looking for ways to make the build faster and faster. I started this PR to integrate parallel testing, just like Symfony core has it. For this to work, one database had to be reused (using unique nodes per batch, so tests don't influence eachother). Parallel testing itself didn't work out well (tests were just as fast or even slower), but not dropping & setting up the database seems to improve the build speed very much.

Latest master build took 1hr 5min 28sec
The build of this PR took 24min 53sec


.travis.yml
```diff
@@ -13,35 +13,40 @@ sudo: false
...
   - composer require symfony/symfony:${SYMFONY_VERSION} --no-update
-  
-install: composer update --prefer-source $COMPOSER_FLAGS
 
-before_script: vendor/symfony-cmf/testing/bin/travis/phpcr_odm_doctrine_dbal.sh
+install: composer update --prefer-dist $COMPOSER_FLAGS
+
+before_script:
+  - vendor/symfony-cmf/testing/bin/travis/phpcr_odm_doctrine_dbal.sh
+  - php vendor/symfony-cmf/testing/bin/console doctrine:phpcr:fixtures:load --fixtures=Tests/Resources/DataFixtures --no-interaction
```
WouterJ: The fixtures are only used by the WebTests, which don't change them. So it's save to only load them once.

.travis.yml
```diff
@@ -13,35 +13,40 @@ sudo: false
...
 
-script: phpunit --coverage-text
+script: phpunit
```
WouterJ: Removed the coverage as it was useless with the parallel testing. I didn't add it back after removing parallel testing, because it reduces testing performance and I never looked at it.
I'm open to add it back /cc @dbu @lsmith77 @dantleech

Tests/Functional/BaseTestCase.php
```diff
@@ -54,6 +54,11 @@ protected function createRoute($path)
      */
     protected function createContent($path = '/test/content')
     {
+        $content = $this->getDm()->find(null, $path);
+        if ($content) {
+            return $content;
+        }
```
WouterJ: Content is never edited, so we can reuse the same node in each test

.travis.yml
dantleech: I don't see a great advantage of having this in CI, I think it just adds noise.

Tests/Functional/Controller/RedirectControllerTest.php
```diff
@@ -18,23 +18,39 @@
...
 
-    public function setUp()
+    /**
+     * Creates the route root.
+     */
+    protected function setUp()
     {
-        parent::setUp();
-        $this->db('PHPCR')->createTestNode();
+        $this->makeSureTestRootExists();
```
dantleech: `initTestNode` ? In anycase I would stick with the existing TestNode terminology instead of TestRoot.

Tests/Functional/Doctrine/Orm/RouteProviderTest.php
```diff
@@ -19,6 +19,7 @@ class RouteProviderTest extends OrmTestCase
 
     public function setUp()
     {
+        $this->markTestSkipped();
```
dantleech: Why skip this now?

Tests/Functional/Doctrine/Phpcr/RedirectRouteTest.php
```diff
@@ -17,15 +17,28 @@
...
-        parent::setUp();
-        $this->db('PHPCR')->createTestNode();
+        $this->makeSureTestRootExists();
         $this->createRoute(self::ROUTE_ROOT);
     }
 
+    /**
+     * Cleans up the created nodes in the database.
+     */
+    protected function tearDown()
```
dantleech: Why don't we move the body of this method into the `BaseTestCase` ? (seems to be a duplicate of the `tearDown` in the RedirectControllerTest)

WouterJ: agreed, applied now

Tests/Functional/Controller/RedirectControllerTest.php
WouterJ: Renamed it.

WouterJ:
Reactivated ORM tests and decided to use only stable deps for most of the tests. This means version resolving is easier and the complete build takes about 15 minutes (if all jobs were executed asynchronously, which isn't the actual case).

Using stable dependencies has the benefit that it's closer to a real installation of the bundle in an application, as the minimum stability is only applied to the root package. One job still tests compatibility with the dev dependencies.

composer.json
```diff
@@ -11,34 +11,33 @@
...
-        "symfony/framework-bundle": "~2.3"
+        "symfony-cmf/routing": "^1.2",
+        "symfony/framework-bundle": "^2.3"
     },
     "require-dev": {
-        "symfony-cmf/core-bundle": "~1.1",
-        "symfony-cmf/testing": "~1.2,>=1.2.7",
-        "doctrine/phpcr-odm": "~1.3",
+        "symfony-cmf/core-bundle": "^1.1",
+        "symfony-cmf/testing": "dev-fix-travis-shell",
```
dbu: testing needs to be merged and tagged and updated here before we merge this PR

composer.json
```diff
@@ -11,34 +11,33 @@
...
+        "doctrine/phpcr-odm": "^1.3@rc",
```
dbu: @lsmith77 do you see a reason not to tag 1.3.0 of phpcr-odm tomorrow?

composer.json
```diff
@@ -11,34 +11,33 @@
...
         "matthiasnoback/symfony-config-test": "0.*",
         "phpunit/php-code-coverage": "@stable",
-        "sonata-project/doctrine-phpcr-admin-bundle": "~1.1",
-        "symfony/monolog-bundle": "~2.3",
-        "doctrine/orm": "~2.3",
-        "doctrine/data-fixtures": "1.*,>=1.0.0-alpha3"
+        "sonata-project/doctrine-phpcr-admin-bundle": "^1.1",
+        "symfony/monolog-bundle": "^2.3",
+        "doctrine/orm": "^2.3",
+        "doctrine/doctrine-bundle": "@dev",
```
dbu: do we miss a stable release of the base doctrine bundle?

dbu:
wow, great job!

can you merge and release the testing fix, and i see about tagging phpcr-odm stable. then we can squash commits here and merge.

.travis.yml
lsmith77: yeah .. I think we only look at failures and never at code coverage on travis

composer.json
WouterJ: The latest stable version has a big bug when creating a SQLite database. This is fixed, but a new version has not yet been released.

WouterJ:
Updated to use Testing 1.3 and squashed commits

composer.json
```diff
@@ -11,34 +11,33 @@
...
-        "symfony/framework-bundle": "~2.3"
+        "symfony-cmf/routing": "^1.2",
+        "symfony/framework-bundle": "^2.3"
     },
     "require-dev": {
-        "symfony-cmf/core-bundle": "~1.1",
-        "symfony-cmf/testing": "~1.2,>=1.2.7",
-        "doctrine/phpcr-odm": "~1.3",
+        "symfony-cmf/core-bundle": "^1.1",
+        "symfony-cmf/testing": "dev-add_methods",
```
WouterJ: On hold till https://github.com/symfony-cmf/Testing/pull/118 is merged

PR #323, opened 2015-12-20
[WIP] Symfony 3

killerwolf:


composer.json
```diff
@@ -14,18 +14,17 @@
     "minimum-stability": "dev",
     "prefer-stable": false,
     "require": {
-        "php": "^5.3.9|^7.0",
+        "php": "^5.5.9|^7.0",
         "symfony-cmf/routing": "~1.2",
-        "symfony/framework-bundle": "~2.3"
+        "symfony/framework-bundle": "~2.8|~3.0"
```
WouterJ: What makes use requiring 2.8 instead of, let's say, 2.7?

composer.json
```diff
@@ -14,18 +14,17 @@
...
     "require-dev": {
-        "symfony-cmf/core-bundle": "~1.1",
+        "symfony-cmf/core-bundle": "dev-sf_3@dev",
         "symfony-cmf/testing": "~1.2,>=1.2.7",
         "doctrine/phpcr-odm": "~1.3",
         "matthiasnoback/symfony-dependency-injection-test": "~0.6",
-        "matthiasnoback/symfony-config-test": "0.*",
+        "matthiasnoback/symfony-config-test": "^1.3.1",
         "phpunit/php-code-coverage": "@stable",
-        "sonata-project/doctrine-phpcr-admin-bundle": "~1.1",
```
WouterJ: this should not be removed. If we update this to `^1.1|^2.0`, I think we'll have Symfony 3 support.

Tests/Unit/Validator/Constraints/RouteDefaultsValidatorTest.php
```diff
@@ -45,7 +45,7 @@ public function setUp()
     {
         $this->controllerResolver = $this->getMock('Symfony\Component\HttpKernel\Controller\ControllerResolverInterface');
         $this->templating = $this->getMock('Symfony\Bundle\FrameworkBundle\Templating\EngineInterface');
-        $this->context = $this->getMock('Symfony\Component\Validator\ExecutionContextInterface');
+        $this->context = $this->getMock('Symfony\Component\Validator\Context\ExecutionContextInterface');
```
WouterJ: if we use a condition for this, we can keep support for Symfony 2.3 I think.

```php
$this->context = $this->getMock(
    class_exists('Symfony\Component\Validator\Context\ExecutionContextInterface')
    ? 'Symfony\Component\Validator\Context\ExecutionContextInterface'
    : 'Symfony\Component\Validator\ExecutionContextInterface'
);
```

killerwolf:
the cmf_tree thing

    cmf_tree:
    resource: .
    type: cmf_tree

generates all the errors about `Unable to load resource "."`

commenting it helped me pass some functional tests

composer.json
dbu: seems to be fixed

dbu:
wouter, is the tree bundle tested with symfony 3? or are we missing something there?

WouterJ:
Problem is that TreeBrowserBundle 2.0-dev is installed, which does not have any routes (routing is done by ResourceRestBundle).

To fix this, `minimum-stability: dev` should be removed from `composer.json`.

WouterJ:
Hmm, actually, we need to update the configuration and the stable versions do not support Symfony 3.

As configuration should no longer live in this bundle, we should copy the relevant test config files from symfony-cmf/Testing to this bundle directly and make sure the configuration works with TreeBrowserBundle 2.0-dev.

dbu:
at that point, i think we should try to release a symfony 2.8 version of RoutingBundle asap, without support for symfony 3. and in this PR ditch support for symfony < 2.8 to move forward. otherwise we depend on tree and resource that are not stable and maybe not ready for release for another month or more.

killerwolf:
So...too early for a symfony3 support.
@dbu would you clarify your last comment ? we 'freeze' this PR (and branch too)..and extend current support to sf2.8..and then work on exclusive suport for sf2.8, right ?

dbu:
yes, that would be my idea to move forward. because supporting 2.3 and 3.0 in the same code seems like a going down a rabbithole and ending up with too many moving targets to hit. but for 2.8, there is probably very little missing, i really hope to tag it in the next few days.

if this branch is at the moment synced to master, you could also start to work on it. if we need to change anything in master, i can take care of fixing eventual conflicts - i don't expect many. see #324

dbu:
most of this is now merged as part of #326

now we can work on refactoring things to work with symfony 2.8 and 3.x, dropping the legacy support.

hacfi:
Do I understand correctly that Symfony 3.0 support is blocked by the dependencies not having stable versions supporting it? I have some limited time to work on Symfony 3.0 compatibility as I need it urgently.

dbu:
i think not even, its mostly that this bundle itself is using some deprecated things. our goal is to release 1.3 in the next few days with 2.8 support, so that people on 2.8 can update, and all the features and fixes become available.

if you want to work on symfony 3 support, please start by creating a branch where you do not set the DEPRECATIONS_HELPER env variable and look at the deprecations. we need to refactor code to not trigger the warnings anymore. set the version requirement to 2.8.*, that version does not need to support older versions of symfony (to keep things simple, and avoid tons of compatibility hacks). you can open a PR as soon as you want so we can discuss issues and fixes.

hacfi:
Thanks David! Will give it a try tonight

dbu:
great!

hacfi:
@dbu https://github.com/symfony-cmf/RoutingBundle/blob/master/Tests/Unit/Routing/DynamicRouterTest.php#L61-L72 calls the deprecated match method: https://github.com/symfony-cmf/Routing/blob/master/DynamicRouter.php#L218
Should I leave it untouched or remove it as well (the test + the method in the routing library)?

hacfi:
@dbu I have to give up for now. It’s quite time-consuming to find out the source of the deprecated call. After several hours I’m down to 77 remaining deprecation notices: https://gist.github.com/hacfi/500098dea33bfa078b74
Will have to clean up my fixes and gather the questions I have (especially regarding BC for sonata admin, updating symfony-cmf/testing etc.). Could be a couple of weeks until I have some actual free time to work on this. If someone starts working on Symfony 3.0 ping me and I might be able to add my fixes.

dbu:
@hacfi thanks for the update. can you do a new WIP pull request with your changes? i am sure it will be helpful for whoever picks this up again.

ftr: i would say we can remove that test.

hacfi:
@dbu Sorry for the late follow-up! I created #331

PR #325, opened 2015-12-22
Symfony 2.8 support

killerwolf:


killerwolf:
should we then just try to remove deprecation notices/warnings then ? which will be hard because some dev dependencies relies on deprecated features methods/classes.  So upgrading is mandatory.

composer.json
```diff
@@ -23,9 +23,9 @@
         "symfony-cmf/testing": "~1.2,>=1.2.7",
         "doctrine/phpcr-odm": "~1.3",
         "matthiasnoback/symfony-dependency-injection-test": "~0.6",
-        "matthiasnoback/symfony-config-test": "0.*",
+        "matthiasnoback/symfony-config-test": "^1.3.1",
         "phpunit/php-code-coverage": "@stable",
-        "sonata-project/doctrine-phpcr-admin-bundle": "~1.1",
+        "sonata-project/doctrine-phpcr-admin-bundle": "^1.1|^2.0",
```
dbu: we should not allow sonata admin 2.0 as per https://github.com/symfony-cmf/RoutingBundle/pull/325#issuecomment-166959786

dbu:
i actually don't mind the deprecations for this branch. i am fine just setting the deprecation helper to not treat them as errors. we need to fix them for the version that supports 3.0 but for 2.8 i would rather have something simple that works with 2.3-2.8 and then we move on with the 3.0 upgrade and don't need workarounds for older symfony versions.

dbu:
continuing in #326

PR #328, opened 2015-12-29
Improve test set-up

WouterJ:


.travis.yml
```diff
@@ -12,33 +12,32 @@ sudo: false
 
 cache:
   directories:
-    - $HOME/.composer/cache
+    - $HOME/.composer/cache/files
```
WouterJ: only files is usefull to cache, the other directories contain files that frequently (every hour) change. Caching them will only reduce test speed.

.travis.yml
```diff
@@ -12,33 +12,32 @@ sudo: false
...
-      env: SYMFONY_VERSION=2.3.* SYMFONY_DEPRECATIONS_HELPER=weak
+      env: SYMFONY_VERSION=2.3.*
     - php: 5.6
-      env: SYMFONY_VERSION=2.8.* SYMFONY_DEPRECATIONS_HELPER=309
+      env: SYMFONY_VERSION=2.7.*
     - php: 5.6
-      env: SYMFONY_VERSION=3.0.*
-  allow_failures:
-    - env: SYMFONY_VERSION=3.0.*
+      env: DEPS=dev
```
WouterJ: all builds now run with stable versions. This will (a) speed up composer a lot and (b) simulate a normal install of the bundle in an application. Only one build will install the dev dependencies.

.travis.yml
```diff
@@ -12,33 +12,32 @@ sudo: false
...
-      env: SYMFONY_VERSION=2.3.* COMPOSER_FLAGS="--prefer-lowest"
+      env: COMPOSER_FLAGS="--prefer-lowest"
     - php: 5.6
-      env: SYMFONY_VERSION=2.3.* SYMFONY_DEPRECATIONS_HELPER=weak
+      env: SYMFONY_VERSION=2.3.*
     - php: 5.6
-      env: SYMFONY_VERSION=2.8.* SYMFONY_DEPRECATIONS_HELPER=309
+      env: SYMFONY_VERSION=2.7.*
     - php: 5.6
-      env: SYMFONY_VERSION=3.0.*
```
WouterJ: the 3.0 build was removed as master does not support 3.0 and we have no plan to support it. Still executing the job only means reducing test speed of the complete CMF org (as an org is only allowed to have 6 concurrent jobs)

WouterJ:
Tests are green now

PR #330, opened 2016-01-09
do not fail on non-default naming strategies

### LINKED ISSUE #309: SchemaException when using symfony 2.7 and persistence.orm.enabled:tue
after composer require symfony-cmf/routing-bundle

config.yml:

```
cmf_routing:
    chain:
        routers_by_id:
            cmf_routing.dynamic_router: 200
            router.default:             100
    dynamic:
            persistence:
                orm:
                    enabled:      true
                  
```

and register in AppKernel.php

on command line
```
php app/console doctrine:schema:update --dump-sql
```

will throw an Exception in symfony 2.7 with orm

```
Doctrine\DBAL\Schema\SchemaException]                              
  There is no column with name 'staticPrefix' on table 'orm_routes'.  
                                                                    
```

the table is not even created yet.

When creating a new symfony demo project (which uses sqlite) it works fine.
But not with pdo_mysql

dbu:
hardcode column names that are used in index definitions to not fail on non-default naming strategies.

as far as i can tell, this should be no BC break, as non-default naming strategies previously must have lead to an error as described in #309. if the user overwrote the mapping, they will not see the change we do here.

fix #309

### DIFF
--- a/Resources/config/doctrine-model/Route.orm.xml
+++ b/Resources/config/doctrine-model/Route.orm.xml
@@ -4,10 +4,10 @@
 
     <mapped-superclass name="Symfony\Cmf\Bundle\RoutingBundle\Model\Route">
         <field name="variablePattern" type="string" nullable="true"/>
-        <field name="staticPrefix" type="string" nullable="true"/>
+        <!-- we hardcode the column name to overwrite column naming strategies as we have to define the index on the column name -->
+        <field name="staticPrefix" type="string" nullable="true" column="staticPrefix"/>
 
         <indexes>
-            <index name="name_idx" columns="name"/>
             <index name="prefix_idx" columns="staticPrefix"/>
         </indexes>
     </mapped-superclass>
--- a/Resources/config/doctrine-orm/Route.orm.xml
+++ b/Resources/config/doctrine-orm/Route.orm.xml
@@ -8,8 +8,13 @@
             <generator strategy="AUTO"/>
         </id>
 
-        <field name="name" type="string" unique="true"/>
+        <!-- we hardcode the column name to overwrite column naming strategies as we have to define the index on the column name -->
+        <field name="name" type="string" unique="true" column="name"/>
         <field name="position" type="integer"/>
+
+        <indexes>
+            <index name="name_idx" columns="name"/>
+        </indexes>
 
     </entity>
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,16 @@
 Changelog
 =========
 
+* **2016-01-09**: [ORM] Hardcoded some column names to match what we index on to avoid
+  issues with non-default `orm.naming_strategy`. It is now safe to use a non-default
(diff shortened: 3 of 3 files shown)


Resources/config/doctrine-model/Route.orm.xml
```diff
@@ -4,10 +4,10 @@
 
     <mapped-superclass name="Symfony\Cmf\Bundle\RoutingBundle\Model\Route">
         <field name="variablePattern" type="string" nullable="true"/>
-        <field name="staticPrefix" type="string" nullable="true"/>
+        <!-- we hardcode the column name to overwrite column naming strategies as we have to define the index on the column name -->
+        <field name="staticPrefix" type="string" nullable="true" column="staticPrefix"/>
 
         <indexes>
-            <index name="name_idx" columns="name"/>
```
dbu: the name field is not defined in the Model class, only in the Doctrine\ORM class. moved it down - i hope it is possible to add indexes from several mapping files.

Resources/config/doctrine-orm/Route.orm.xml
```diff
@@ -8,9 +8,14 @@
             <generator strategy="AUTO"/>
         </id>
 
-        <field name="name" type="string" unique="true"/>
+        <!-- we hardcode the column name to overwrite column naming strategies as we have to define the index on the column name -->
+        <field name="name" type="string" unique="true" column="name"/>
```
dbu: not sure what strategy could make this fail, but better be safe.

dbu:
@ocramius is this what you meant by "prefixing" in https://github.com/symfony-cmf/RoutingBundle/issues/309#issuecomment-170225637 ?

dbu:
@Nielsb85 @steve-todorov does this look like it would fix the issue you described in #309?

Ocramius:
@dbu I sadly don't know this config format, and therefore can't answer your question :-(

dbu:
@Ocramius its the same for all mappings, e.g. annotations has this: https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Mapping/Index.php . they are all based on the column name and no way to specify the field of the entity and have the column looked up from the evaluated metadata.

dbu:
@Nilsb85 would you mind checking if this fix works for you?

Nielsb85:
@dbu will check asap. I hope somewhere beginning of next week.
thanks.

dbu:
thanks. i merge this now so that we can tag RC1. looking forward to your feedback, we can still fix things until next week and then release a stable 1.4.0

Nielsb85:
@dbu the schema is updated successfully. There is a minor inconsistency now with the column name staticPrefix, which is the only column that is now in camel case while the rest is using underscore. But it does fix the problem I described.

dbu:
thanks for the follow-up. yes, that can't be avoided without overwritting the mapping in your project, unfortunately. at least the configuration setting now does not break the application anymore.

PR #331, opened 2016-01-30
[WIP] Symfony3

hacfi:

PR #332, opened 2016-02-17
Symfony3 compatibility

hacfi:
- allow Symfony ~3.0
- bump Symfony dependency to ~2.7
- specify correct Sonata dependencies
- use parentDocument in Admin classes
- update route_basepaths in tests
- make StyleCI happy

See #331

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ sudo: false
 cache:
   directories:
     - $HOME/.composer/cache/files
- 
+
 env:
   global:
     - SYMFONY_DEPRECATIONS_HELPER=322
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -21,11 +21,9 @@ env:
 matrix:
   include:
     - php: 5.6
-      env: DEPS=dev
+      env: DEPS=dev SYMFONY_VERSION=2.8.*
```
WouterJ: May I do one last try? :)

```
DEPS=dev SYMFONY_VERSION="3.0.x-dev as 2.8.x-dev"
```

dbu: or i just merge and we can do further PRs

dbu:
thanks a lot @hacfi 
i merged this so that we can do independent PRs to work on the build matrix and getting rid of deprecated / BC things.

WouterJ:
We now have a `symfony3` and `symfony_3` branch, I think the one with the underscore has to be removed?

dbu:
ooops. sorry, did not notice that one (in my defense, there is no open
PR for it, either).

most of symfony_3 is about adding BC for the forms, which is not needed
now that we never support 2.3 and 3.0 with the same version. so i guess
we can simply delete that. however, the change
00c9db42927e057303ef2fbb9c70c3d610910a1b might be relevant, or not? it
actually should go into the 1.4 version, it is not about the symfony
version.

PR #339, opened 2016-02-28
fix cs styleci.io

andrey1s:
| Q             | A
| ------------- | ---
| Bug fix?      | [no]
| New feature?  | [no]
| BC breaks?    | [no]
| Deprecations? | [no]
| Tests pass?   | [yes]
| Fixed tickets | []
| License       | MIT
| Doc PR        | []

fix code style https://styleci.io/analyses/qgOEDQ


dbu:
thanks!
it looks like the deps build has a tendency to hang on travis-ci. i guess there are too many options. but this change is not the cause for that, so i merged.

PR #340, opened 2016-03-02
Check Metadata Class info exists

gerfigna:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT

Sometimes, Metadata ReflectionClass is null and throws:

PHP Fatal error:  Call to a member function getName() on a non-object in vendor/symfony-cmf/routing-bundle/Doctrine/RouteConditionMetadataListener.php on line 52

i.e. I get this error when I try to use doctrine:generate:entities command using orm persistence.

### DIFF
--- a/Doctrine/RouteConditionMetadataListener.php
+++ b/Doctrine/RouteConditionMetadataListener.php
@@ -49,6 +49,11 @@ class RouteConditionMetadataListener implements EventSubscriber
         }
 
         $meta = $eventArgs->getClassMetadata();
+
+        if (null === $meta->getReflectionClass()) {
+            return;
+        }
+
         if ($meta->getReflectionClass()->getName() !== 'Symfony\Component\Routing\Route') {
             return;
         }


Doctrine/RouteConditionMetadataListener.php
```diff
@@ -49,6 +49,11 @@ public function loadClassMetadata(LoadClassMetadataEventArgs $eventArgs)
         }
 
         $meta = $eventArgs->getClassMetadata();
+
+        if (null === $meta->getReflectionClass()) {
```
dbu: is this something that is supposed to ever happen? we would be loading metadata, but not about a class? or in what situation does this problem occur?

gerfigna:
It seems there is a problem with composer update on CI tests

Doctrine/RouteConditionMetadataListener.php
gerfigna: I don't know what is the condition to get the empty medatadata. 
I will investigate a bit more.

I see KnpLabs/DoctrineBehaviors doing the same condition check.

dbu: but did you see this happen in your project?

i am not against doing this, but i would like to understand when this
can happen to avoid hiding a different problem. maybe there is no other
solution and this is normal and can happen, but i am not sure.

WouterJ: When initializing ClassMetadata, the reflection is not automatically initialized. for instance, this is used in the DisconnectedClassMetadataFactory used in `doctrine:generate:entities`.

This allows the `doctrine:generate:entities` command to create the PHP code for not yet exisiting entities (e.g. when [reverse-engineering from a non-doctrine database](http://symfony.com/doc/current/cookbook/doctrine/reverse_engineering.html), which is the main use-case for this command).

WouterJ:
Thank you @gerfigna, it's a nice fix!

Issue #317, opened 2015-12-01
Symfony 3.0 compatibility?


nicolas-grekas:
I don't see this bundle in https://docs.google.com/spreadsheets/d/1GOxJ2lNpLxN12YyZ87Bhsxb0iLAQ7e4UDZbJSly8Sy8/edit#gid=0
Did someone try to make it Symfony compatible?

dbu:
the Routing component is compatible: https://github.com/symfony-cmf/Routing/blob/master/composer.json#L17-L18

so i think it should not be too hard to make the bundle compatible as well. but i am not aware of ongoing work for that at the moment. might be something for the symfony con hackday this saturday.

rande:
Checking now ;)

killerwolf:
+1

killerwolf:
+1 again :)

dbu:
@killerwolf do you want to work on this? create a pull request where you adjust composer.json and the .travis.yml file to allow symfony 3 in addition to 2, then check if something fails and if so, check how to avoid the errors.

killerwolf:
i'll try :) keep all of you updated soon

killerwolf:
In progress here https://github.com/symfony-cmf/RoutingBundle/pull/323
Inputs are welcome

killerwolf:
how to trigger travis build for this PR ?
styleci is yelling at me, is the [] shorthand mandatory ?

killerwolf:
i found why..updated my fork, and every thing seems ok now

dbu:
symfony 2.8 support is now merged in #326. for 3.0, we plan to work on a version 2.0 of this bundle that gets rid of all deprecated things and works with 2.8 and 3.x (see discussion in #324)

WouterJ:
Closing, 2.0-dev is now compatible with Symfony 3.x

PR #347, opened 2016-04-07
Support Symfony 3

WouterJ:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #317 
| License       | MIT
| Doc PR        | -

Created a new branch, so I don't have to rebase a branch that's used by people already.


Tests/Resources/app/config/cmf_routing.phpcr.yml
```diff
@@ -3,4 +3,4 @@ cmf_routing:
         persistence:
             phpcr:
                 enabled: true
-                route_basepath: /test/routing
+                route_basepaths: /test/routing
```
dantleech: should this be an `[ 'array' ]` ? I know that we probably support scalars, but I think its better to treat a plural as a plural.

dbu: not sure. it does not cost us much to convert string to array here for when you have only one path. @WouterJ what do you think is better DX? do we create confusion by allowing single value or is it convenient?

dbu:
hm, why is travis-ci not looking at this?

if this is all it takes, i wonder if we could support symfony 3 with a routing bundle 1.x . should we branch alias to 1.5 and aim to release an 1.5.0 as soon as symfony 3 support is solid?

dbu:
i mean: i don't see any BC breaks in this PR, this is what made me think we could have an 1.x release of this.

WouterJ:
Hmm, indeed, @dbu. However, there is a problem: None of the dev (optional) dependencies support Symfony 3. This means that we can't run the functional and webtests with Symfony 3. It also means that people can't use the admin integration, etc.

If no other CMF bundle supports Symfony 3, does it make sense to support Symfony 3 in this bundle?

dbu:
not sure but i think some people use this without other cmf bundles and
without sonata, so it would be useful to them.

but if we can't run the functional tests, i am not confident to call
things supported. lets be pragmatic and have this go into a 2.0 version
together with the rest then.

WouterJ:
Please note that for Symfony 3 support, people can require `dev-symfony3` as their version constraint btw.

dbu:
can we fix styleci too?

WouterJ:
I now got all tests working on Symfony 3 locally. This requires https://github.com/sonata-project/SonataAdminBundle/pull/3709 and https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/pull/367 to be merged first.

dbu:
https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/pull/367 is merged, is this now ready?

WouterJ:
I'll try to merge most of the open Sf3 PR this evening

WouterJ:
It took a bit longer, but the road to Symfony 3 support is now officially open :)

dbu:
> It took a bit longer, but the road to Symfony 3 support is now
> officially open :)

woohoo, thanks a lot for your efforts wouter!

wachterjohannes:
@dbu will there be progess for 1.x series? in the future?

dbu:
we try to get the 2.0 versions out of the door. at that point we will not spend time on 1.x anymore, we don't have the resources. as far as i can see, 2.0 will however still support symfony 2.8

PR #349, opened 2016-05-02
change controller_by_class configuration to get a action by http meth…

ElectricMaxxx:
| Q             | A
| ------------- | ---
| Bug fix?      | [no]
| New feature?  | [yes]
| BC breaks?    | [no]
| Deprecations? | no]
| Fixed tickets | https://github.com/symfony-cmf/Routing/issues/171

…od configuration

So this ist is first step. Need to do:

* [ ] implement change on routing 
* [ ] add xml schma changes
* [ ] let configuration work

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -83,7 +83,18 @@ class Configuration implements ConfigurationInterface
                         ->end() // controllers_by_type
                         ->arrayNode('controllers_by_class')
                             ->useAttributeAsKey('class')
-                            ->prototype('scalar')->end()
+                            ->prototype('array')
+                                ->beforeNormalization()
+                                    ->ifString()
+                                    ->then(function ($v) {
+                                        return array(array('methods' => ['any'], 'controller' => $v));
+                                    })
+                                ->end()
+                                ->children()
+                                    ->arrayNode('methods')->end()
+                                    ->scalarNode('controller')->end()
+                                ->end()
+                            ->end()
                         ->end() // controllers_by_class
                         ->arrayNode('templates_by_class')
                             ->useAttributeAsKey('class')
--- a/Tests/Resources/Fixtures/config/config.php
+++ b/Tests/Resources/Fixtures/config/config.php
@@ -15,6 +15,13 @@ $container->loadFromExtension('cmf_routing', array(
         ),
         'controllers_by_class' => array(
             'Symfony\Cmf\Bundle\ContentBundle\Document\StaticContent' => 'cmf_content.controller:indexAction',
+            'My\Class' => array(
+                array('methods' => array('put', 'post'), 'controller' => 'service:method'),
+                array('methods' => array('any'), 'controller' => 'service:readMethod'),
+            ),
+            'Other\Class' => array(
+                array('methods' => array('any'), 'controller' => 'service:method'),
+            ),
         ),
         'templates_by_class' => array(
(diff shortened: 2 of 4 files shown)


DependencyInjection/Configuration.php
```diff
@@ -82,8 +82,21 @@ private function addDynamicSection(ArrayNodeDefinition $root)
                             ->prototype('scalar')->end()
                         ->end() // controllers_by_type
                         ->arrayNode('controllers_by_class')
-                            ->useAttributeAsKey('class')
-                            ->prototype('scalar')->end()
+                            ->prototype('array')
+                                ->ignoreExtraKeys(false)
```
ElectricMaxxx: This one was the only chance to get a nested array without an own key, but now i got no validation on deeper key names (value, methods) anymore.

DependencyInjection/CmfRoutingExtension.php
```diff
@@ -319,4 +288,70 @@ public function getNamespace()
...
+                        $serviceDefinitionValues['arguments'][0],
+                        $serviceDefinitionValues['arguments'][1],
+                        [$mapKey => trim($value)]
+                    ),
+                ];
+            }
+        }
+
+        if (0 < count($enhancerMap)) {
+            $container->getDefinition('cmf_routing.enhancer.conditional')->addArgument([]);#$enhancerMap);
```
ElectricMaxxx: HI @WouterJ 
can you give me some help here, or just an advise. @dbu and i cam to the solution of an condional enhancer (symfony-cmf/routing#174) to handle method aware configuration. This one should get an enhancer map as created some lines above. This map unfortunatly contains objects, so i can't inject them as an argurment of a DI service defintion. You have an idea how to pass them in? One idea is to use an array only and build the object based map inside the enhancers constructor, which means to change that one in routing component, you have an other idea?

ElectricMaxxx: https://github.com/symfony-cmf/routing/pull/175 would be my solution.

DependencyInjection/CmfRoutingExtension.php
```diff
@@ -82,7 +87,20 @@ private function setupDynamicRouter(array $config, ContainerBuilder $container,
...
 
         // strip whitespace (XML support)
-        foreach (array('controllers_by_type', 'controllers_by_class', 'templates_by_class', 'route_filters_by_id') as $option) {
+        foreach (array('controllers_by_type', 'controllers_by_class', 'templates_by_class') as $option) {
+            $config[$option] = array_map(function ($items) {
+                foreach ($items as $key => $item) {
+                    if (is_array($item)) {
+                        $items[$key]['value'] = trim($item['value']);
+                    } else {
+                        $items[$key] = trim($item);
```
dbu: @WouterJ do you know why we need to trim here? xml configuration ease of use?

ElectricMaxxx: There also is an WhiteListTest. I think is a point of the controller resolver, that tries to get the controller as a service, isn't it?

dbu: its indeed the xml configuration. we have tests making sure that controllers with whitespace around them do work. surprised though that this is not in the Configuration class instead.

ElectricMaxxx: by the normalizer, right?

Am 21.05.2016 um 09:54 schrieb David Buchmann:

dbu:
ok if we postpone this to later than 2.0? as long as its BC, it can be in 2.1 or whenever we are ready

PR #353, opened 2016-05-21
WIP conditional enhancers

dbu:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

I spent quite some time on this until i realized what you tried to explain to me @ElectricMaxxx .

before, we had e.g. one enhancer for the controller by class map. now we would need a conditional enhancer per entry in that map with a field by class enhancer per method. this is indeed super inefficient and cumbersome.

an easy solution could be to swap the configuration. if we had this:

```yaml
controllers_by_class:
    - 
        methods: put
        map:
            My\Class: service:method
    - 
        methods: any
        map:
            My\Class: service:readMethod
            Other\Class: service:method    
```

this is a bit less readable and a bit confusing, but would reduce the overhead tremendously.

i pondered rearranging the mapping to group it by method, but with the "any" entries, this could change behaviour if we have this:

```yaml
controllers_by_class:
    My\ChildClass:
        - 
            methods: [any]
            value: cmf_content.controller:indexAction
    My\ParentClass:
        -
            methods: [put, post]
            value: service:method
```

if we first execute the put, post matcher, we end up with service:method for the child class, instead of the any that was specific to ChildClass. with the new configuration, this will be explicit and can be noticed and adressed by adding specific entries for ChildClass.

### DIFF
--- a/DependencyInjection/CmfRoutingExtension.php
+++ b/DependencyInjection/CmfRoutingExtension.php
@@ -15,7 +15,6 @@ use Symfony\Cmf\Component\Routing\Enhancer\ConditionalEnhancer;
 use Symfony\Cmf\Component\Routing\Enhancer\FieldByClassEnhancer;
 use Symfony\Cmf\Component\Routing\Enhancer\FieldMapEnhancer;
 use Symfony\Component\DependencyInjection\Definition;
-use Symfony\Component\HttpFoundation\RequestMatcher;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
@@ -307,13 +306,24 @@ class CmfRoutingExtension extends Extension
      */
     private function loadConditionalEnhancer(array $config, ContainerBuilder $container)
     {
-        $enhancerMap = [];
-        $definitions = [];
-        $configurationPriorities = [
-            'templates_by_class' => ['class' => FieldByClassEnhancer::class, 'arguments' => ['_content', '_template'], 'priority' => 40],
-            'controllers_by_class' => ['class' => FieldByClassEnhancer::class, 'arguments' => ['_content', '_controller'], 'priority' => 50],
-            'controllers_by_type' => ['class' => FieldMapEnhancer::class, 'arguments' => ['type', '_controller'], 'priority' => 60],
-        ];
+        $definitions = array();
+        $configurationPriorities = array(
+            'templates_by_class' => array(
+                'class' => FieldByClassEnhancer::class,
+                'arguments' => array('_content', '_template'),
+                'priority' => 40
+            ),
+            'controllers_by_class' => array(
+                'class' => FieldByClassEnhancer::class,
+                'arguments' => array('_content', '_controller'),
+                'priority' => 50
+            ),
+            'controllers_by_type' => array(
+                'class' => FieldMapEnhancer::class,
+                'arguments' => array('type', '_controller'),
+                'priority' => 60
+            ),
+        );
 
 
         if (count($config['templates_by_class']) > 0) {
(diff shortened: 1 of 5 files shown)


DependencyInjection/CmfRoutingExtension.php
```diff
@@ -371,14 +384,32 @@ private function loadConditionalEnhancer(array $config, ContainerBuilder $contai
...
+            $enhancer->addTag('dynamic_router_route_enhancer', array('priority' => $priority));
+
+            return $enhancer;
+        }
+
+        $enhancerMap = [];
+        foreach ($maps as $map) {
+            $matcher = new Definition('Symfony\Component\HttpFoundation\RequestMatcher');
+            $matcher->setArguments(array(null, null, $map['methods']));
+            $enhancer = new Definition($className);
```
ElectricMaxxx: ohhhh .... i am never had that in mind ... sorry.

DependencyInjection/CmfRoutingExtension.php
```diff
@@ -371,14 +384,32 @@ private function loadConditionalEnhancer(array $config, ContainerBuilder $contai
         $container->addDefinitions($definitions);
     }
 
-    private function buildConditionalEnhancerDefintion($className, $source, $target, $map, $methods, $priority)
+    private function buildConditionalEnhancerDefintion($className, $source, $target, $maps, $priority)
     {
+        var_dump($maps);throw new \Exception;
```
ElectricMaxxx: is that the new `var_dump() or die()`?

DependencyInjection/Configuration.php
```diff
@@ -200,7 +208,7 @@ public function methodAwareValues($name)
                 ->beforeNormalization()
                     ->ifString()
                     ->then(function ($v) {
-                        return [['methods' => ['any'], 'value' => $v]];
+                        return array('value' => $v);
```
ElectricMaxxx: not setting `methods => ['any'] is the way to recognize old definitions?

DependencyInjection/CmfRoutingExtension.php
```diff
@@ -371,14 +384,32 @@ private function loadConditionalEnhancer(array $config, ContainerBuilder $contai
...
+        if (1 === count($maps) && !array_key_exists('methods', reset($maps))) {
```
ElectricMaxxx: we should allow manually setting `any` as an method value instead of the old defintion too. Can be confusing else. 
So

- old configuration => stays old (no method) => can be merged in one group per configuration block
- new configuration with any in methods = old configuration => can go in the same block

DependencyInjection/CmfRoutingExtension.php
```diff
@@ -371,14 +384,32 @@ private function loadConditionalEnhancer(array $config, ContainerBuilder $contai
         $container->addDefinitions($definitions);
     }
 
-    private function buildConditionalEnhancerDefintion($className, $source, $target, $map, $methods, $priority)
```
ElectricMaxxx: should be changed for the other call too.

DependencyInjection/CmfRoutingExtension.php
dbu: the PR does say WIP at least :-)

i needed to figure out what was going on to understand i have a total misconception.

DependencyInjection/CmfRoutingExtension.php
dbu: my idea is to avoid extra wrappers when they are not needed. but i guess you are right that methods=any should also be this case.

dbu:
@ElectricMaxxx did you read the description as well? i realized that my idea really does not work out like i thought. basically we either need to change the configuration format to swap methods to the outer loop, or we need some sort of mixed enhancer that uses the matcher together with the actual operation.

just thought, we could also use a matcher to decide if a specific field is present and represents a class, or the value. then we would not even need the separate FieldMapEnhancer and FieldByClassEnhancer but have just one MapEnhancer that uses a list of specific FieldClassMatcher / FieldMatcher and the field value if match. then we could drop the ConditionalEnhancer again. do you see what i mean? i can try to write that out tonight...

dbu:
@ElectricMaxxx ping

ElectricMaxxx:
actually we have a chain with all enhancers and each ehancer is asked to enhance. Each enahncer does a pre-selection (early outs) if it stll can do something and is responsible. Even when that means we have to change the enhancers itself i would propose to move that early outs into a match() method on the enhancer. So we can loop throug them all, use the RequestMatcher for an outer match and the enhancerMatcher for an enhancer match. So we can lopp throug them all and won't change anything or we can group the enhancer based on source/target means their responsibility.

ElectricMaxxx:
@dbu ping you have some time to talk about that theme?

PR #284, opened 2015-01-19
Use the tab features of SonataAdminBundle 2.3

WouterJ:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

This should also work under SonataAdminBundle <2.3. This'll use both groups (columns) and tabs in the admin panel, improving the UX a lot :)

I've only reviewed it as parent of the SimpleCms admin, that's why I tagged it as WIP.

This requires https://github.com/sonata-project/SonataAdminBundle/pull/2615 to be merged.

See https://github.com/symfony-cmf/SimpleCmsBundle/pull/131 for an example.

PR #355, opened 2016-06-11
[WIP] Move sonata admin to separate bundle

ElectricMaxxx:
| Q | A |
| --- | --- |
| Bug fix? | [no] |
| New feature? | [no] |
| BC breaks? | [yes] |
| Tests pass? | [yes] |
| License | MIT |

Todo:
- [x] implement form types
- [x] implement DTO
- [x] implement converter for DTO
- [ ] green on [sandbox](https://github.com/symfony-cmf/cmf-sandbox/pull/332)
- [ ] green on [sonata-admin-integration])


WouterJ:
All these removals make me worry a bit about the hard upgrade path to CMF 2. What about making the configuration and services backwards compatible? (by requiring the SonataAdminIntegrationBundle and using those classes in the service definitions and using prepended extensions).

We can then deprecate the old services and configuration and remove them in CMF 3.

ElectricMaxxx:
@WouteJ You should have a look at the sanbox PR. The changes are on configuration only.

Admin/Extension/RouteReferrersExtension.php
```diff
@@ -1,44 +0,0 @@
...
-                'translation_domain' => 'CmfRoutingBundle',
-            ))
-            ->add(
-                'routes',
-                CollectionType::class,
-                array(),
-                array(
-                    'edit' => 'inline',
-                    'inline' => 'table',
-                )
```
dbu: @WouterJ should we add a form type for things like this as well, or is that overengineering?

Admin/RedirectRouteAdmin.php
```diff
@@ -1,73 +0,0 @@
...
-
-    protected function configureFormFields(FormMapper $formMapper)
-    {
-        $formMapper
-            ->with('form.group_general')
-                ->add('parentDocument', TreeModelType::class, array('choice_list' => array(), 'select_root_node' => true, 'root_node' => $this->routeRoot))
-                ->add('name', TextType::class)
-                ->add('routeName', TextType::class, array('required' => false))
-                ->add('uri', TextType::class, array('required' => false))
-                ->add('routeTarget', TreeModelType::class, array('choice_list' => array(), 'required' => false, 'root_node' => $this->routeRoot))
```
dbu: we should convert this to a formtype in this bundle that we can then use in the admin integration bundle

Admin/RouteAdmin.php
```diff
@@ -1,227 +0,0 @@
...
-
-        if (null === $this->getParentFieldDescription()) {
-            $formMapper
-                ->with('form.group_general', array(
-                    'translation_domain' => 'CmfRoutingBundle',
-                ))
-                    ->add('content', TreeModelType::class, array('choice_list' => array(), 'required' => false, 'root_node' => $this->contentRoot))
-                ->end()
-                ->with('form.group_advanced', array(
-                    'translation_domain' => 'CmfRoutingBundle',
```
dbu: we should have a form type for the general things and one for the advanced part.

dbu:
i would prefer to get rid of the dependency on sonata admin asap. what if we let the admin integration bundle provide the old service names as alias (but deprecating them from the beginning)? then people that did not extend admin classes only need to require + load the new bundle (they need to load it even when we depend on it, so there is some change involved).

Admin/Extension/RouteReferrersExtension.php
ElectricMaxxx: Will put it into Form types later. Just whanted to get it working first.

dbu: sure, no worries. just wanted we don't forget to do it :-)

thanks a lot for getting this started!

dbu:
@ElectricMaxxx can you maybe check #284 and take whatever is still relevant into the admin integration bundle as well?

WouterJ:
I wouldn't do that now, @dbu. Let's first move the current code to integration and put form types here and then I'll try to get the tabs working again.

ElectricMaxxx:
:+1: for @WouterJ lets do behind. So we will have a better comparisson too. And i think working on sonata will ne easier then and we can build a more common admin interface then.

ElectricMaxxx:
@dbu i do need some insights here: is there some magig on the `parentDocument` Field? I see (while debuggin) that a field with name `parentDocument` is added, but the tes says no. and there i see a field with name `parent`. That is on both Admins. feels really strange.

dbu:
we have the [Route::parent](https://github.com/symfony-cmf/routing-bundle/blob/master/Doctrine/Phpcr/Route.php#L34) field but deprecated using setParent / getParent in favor of setParentDocument / getParentDocument. the form layer (and thus sonata) does not care about private property like doctrine does. it uses property access, so if you say parentDocument, it looks for get/setParentDocument or a public property parentDocument. it should find the setter and getter for parentDocument.

Admin/Extension/RouteReferrersExtension.php
ElectricMaxxx: @dbu @WouterJ can you give me a hand on replacing that by a custom form type. I am ablte to create a form type with fields like we have in the admin, but how should i add them here for several fields instead of a single field as i know it?

Form/Type/RedirectRoute.php
```diff
@@ -0,0 +1,57 @@
...
+use Sonata\DoctrinePHPCRAdminBundle\Form\Type\TreeModelType;
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\TextType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * @author Maximilian Berghoff <Maximilian.Berghoff@mayflower.de>
+ */
+class RedirectRoute extends AbstractType
```
WouterJ: RedirectRouteType

Form/Type/RouteAdvanced.php
```diff
@@ -0,0 +1,49 @@
...
+use Sonata\CoreBundle\Form\Type\ImmutableArrayType;
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\TextType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * @author Maximilian Berghoff <Maximilian.Berghoff@mayflower.de>
+ */
+class RouteAdvanced extends AbstractType
```
WouterJ: RouteAdvancedType

Form/Type/RedirectRoute.php
ElectricMaxxx: Isn't that a duplication when living in a folder called `Type` and in a
namespace called `Type`?

Admin/Extension/RouteReferrersExtension.php
ElectricMaxxx: Got a [hint](http://iltar.nl/) but it will be a more complex work. I really like his idea, used that in a projekt last year. The DTO should stay with the form types inside ther routing-bundle,right?

Form/Type/RedirectRoute.php
dbu: it seems to be a symfony convention nonetheless. with a `use` statement
you don't see the Type part anymore - not sure if that is the only reason.

Admin/Extension/RouteReferrersExtension.php
dbu: the routing bundle should contain all that one needs to use the form.

Form/Type/RedirectRoute.php
ElectricMaxxx: Ok. I will implement the DTO stuff. And put the convert into the AdminClass first. Maybe we will find an abstraction for that later.

Admin/Extension/RouteReferrersExtension.php
ElectricMaxxx: :+1:

Form/Type/RedirectRoute.php
ElectricMaxxx: Found an other reason: using the `::class` notation for the form type usage.

Admin/Extension/RouteReferrersExtension.php
fbourigault: What are the DTOs solving here?

ElectricMaxxx:
@*fbourigault*
Not sure if the DTO are solving something. Saying we would do same as
you did on contentbundle would make a lot of stuff easier.
So the extraction is still ready on that PR. Even the insert into
sonata-admin-integration bundle works. What is missing here, is the
reusable form type. What do you think how to solve it in a good way?

fbourigault:
I'm doing the work for CoreBundle which looks like a lot easier. RoutingBundle has coupling on FormType that need to be solved. See #361.

wouterj:
@ElectricMaxxx you should do a rebase of master, everything works fine in master...

ElectricMaxxx:
Yes right. Got it working again. Just need some more tests.

ElectricMaxxx:
I think i am on a good way with that, FormTypes, DTOs and its converter are working fine. just have an testing issue in the functional tests:

```
[PHPCR] Error when initializing dbal: 
                                                                                                                                       
  [Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException]                                                           
  The service "cmf_routing.route_general_form_type" has a dependency on a non-existent service "sonata.admin.manager.doctrine_phpcr".  
 ```
any ideas?

ElectricMaxxx:
@wouterj is again the testing componant that doesn't allow me to set the right bundle sets like it is done for the web tests in other bundles/components?

dbu:
moved to milestone 2.1 because this PR also works on form types

ElectricMaxxx:
all right, i will try to extract/cherry pick the usefull commits from here, cause i had been on a working state before starting with the form types

ElectricMaxxx:
Closed due to #369 will do the form type stuff stand alone.

PR #356, opened 2016-06-18
Removed *.class parameters

WouterJ:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+2.0.0 (unreleased)
+------------------
+
+ * **2016-06-18**: [BC BREAK] Removed all `*.class` parameters.
+
 1.4.0
 -----
 
--- a/Resources/config/admin-phpcr.xml
+++ b/Resources/config/admin-phpcr.xml
@@ -4,21 +4,12 @@
            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
            xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
 
-    <parameters>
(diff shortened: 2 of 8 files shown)


gonzalovilaseca:
Why were the class parameters removed? 
What would the best practice be to override the `ChainRouter` class? Extend the bundle and add the service definition with out custom class?

dbu:
> Why were the class parameters removed?
> What would the best practice be to override the |ChainRouter| class?
> Extend the bundle and add the service definition with out custom class?

this pattern got deprecated. if we would rename a class parameter, it
would randomly no longer match but not give you any error. the best
practice is to do a compiler pass that loads the service definition and
changes the class - and adjusts the service definition if needed.

why are you using a custom chain router? if its a general use case,
maybe we can also improve the ChainRouter provided by this bundle.

gonzalovilaseca:
I don't think it's a general use case: we have url prefixes such as `es`, `us`, `rw` for all our routes, so that all of them are regionalised (and that dictates prices, etc.), when we generate a route, we add this prefix to it.

dbu:
yes, this sounds like a special solution.

note: this is usually done in the specific router, not ChainRouter.
DynamicRouter supports locale prefixes, and for the core symfony router,
bundles to translate the URLs usually can do this as well.

gonzalovilaseca:
Thanks for the suggestion!
We tried that approach, but we couldn't regionalise some bundles, so we moved it here as we regionalise all routes from all bundles. A request listener removes the prefix, and then the request is parsed normally, then we add the prefix when generating any url.

dbu:
okay, thanks for explaining the use case. not something we need to move to
cmf, but interesting solution.

Am 13.03.2017 5:42 nachm. schrieb "Gonzalo Vilaseca" <
notifications@github.com>:

PR #359, opened 2016-07-05
declare conflict with phpcr-odm 1.*

dbu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

since removing the getParent method from the Route document, we conflict with phpcr-odm 1.* that had this in its interface

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,6 +43,9 @@
 
         "sonata-project/doctrine-phpcr-admin-bundle": "To provide an admin interface for the PHPCR ODM documents (^1.1)"
     },
+    "conflict": {
+        "doctrine/phpcr-odm": "<2.0"
+    },
     "autoload": {
         "psr-4": {
             "Symfony\\Cmf\\Bundle\\RoutingBundle\\": ""


WouterJ:
Do we really want this? We don't do this for other optional requirements (e.g. this package conflicts with `sonata-project/doctrine-phpcr-admin-bundle` `<1.2` as well).

Also, conflict rules aren't taken into account during version resolving and are only checked when versions are resolved IIRC. I don't like using conflict too much

dbu:
i installed this accidentally together with phpcr-odm 1.3 and was
confused why i suddenly get the error that the route has abstract
methods in it. as we don't depend on phpcr-odm directly, i think the
conflicts would be useful to avoid installing an incompatible combination.

dbu:
i discovered the problem when running a composer update in the sandbox, without any other arguments. because the sandbox depends on phpcr-odm ^1.3 but we now conflict with it. i'd rather add the sonata constraint as well than not add any constraints at all.

PR #363, opened 2016-10-07
Add option to use custom Route class in ORM

ahilles107:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | [symfony-cmf/symfony-cmf-docs#779](https://github.com/symfony-cmf/symfony-cmf-docs/pull/779)

As PHPCR easily allows you to add custom properties (and own class) to Route, this functionality ix missing in ORM implementation. This PR replaces hardcoded Route class name with configuration parameter and register doctrine mapping config pass for only needed directories. 

In case of custom Route class all what you need to do is adding `route_class` to orm config:
```
cmf_routing:
    ...
    dynamic:
        persistence:
            orm:
                enabled: true
                route_class: "SWP\\Bundle\\ContentBundle\\Doctrine\\ORM\\Route"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 Changelog
 =========
 
+ * **2016-10-10**: [ORM] Add option to use custom Route class in ORM
  * **2016-06-18**: [BC BREAK] Removed all `*.class` parameters.
  * **2016-06-18**: [BC BREAK] Removed the `cmf_routing.dynamic.persistence.phpcr.route_basepath`
    setting.
--- a/CmfRoutingBundle.php
+++ b/CmfRoutingBundle.php
@@ -96,8 +96,19 @@ class CmfRoutingBundle extends Bundle
                     realpath(__DIR__.'/Resources/config/doctrine-orm') => 'Symfony\Cmf\Bundle\RoutingBundle\Doctrine\Orm',
                 ),
                 array('cmf_routing.dynamic.persistence.orm.manager_name'),
-                'cmf_routing.backend_type_orm',
+                'cmf_routing.backend_type_orm_default',
                 array('CmfRoutingBundle' => 'Symfony\Cmf\Bundle\RoutingBundle\Doctrine\Orm')
+            )
+        );
+
+        $container->addCompilerPass(
+            $doctrineOrmCompiler::createXmlMappingDriver(
+                array(
+                    realpath(__DIR__.'/Resources/config/doctrine-model') => 'Symfony\Cmf\Bundle\RoutingBundle\Model',
+                ),
+                array('cmf_routing.dynamic.persistence.orm.manager_name'),
+                'cmf_routing.backend_type_orm_custom',
+                array()
             )
         );
     }
--- a/DependencyInjection/CmfRoutingExtension.php
+++ b/DependencyInjection/CmfRoutingExtension.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Cmf\Bundle\RoutingBundle\DependencyInjection;
 
+use Symfony\Cmf\Bundle\RoutingBundle\Doctrine\Orm\Route;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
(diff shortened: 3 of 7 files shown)


ahilles107:
@dbu @wouterj - what do You think? 

I had to add new CompilerPass which don't include `Resources/config/doctrine-orm` directory to do not create `orm_routes` table automatically (and fix indexes collisions when both classes - default and custom - extend this same base class (with this same mapping configurations).

CmfRoutingBundle.php
```diff
@@ -96,10 +96,21 @@ private function buildOrmCompilerPass(ContainerBuilder $container)
...
-                'cmf_routing.backend_type_orm',
+                'cmf_routing.backend_type_orm_default',
                 array('CmfRoutingBundle' => 'Symfony\Cmf\Bundle\RoutingBundle\Doctrine\Orm')
             )
         );
+
+        $container->addCompilerPass(
+            $doctrineOrmCompiler::createXmlMappingDriver(
+                array(
+                    realpath(__DIR__.'/Resources/config/doctrine-model') => 'Symfony\Cmf\Bundle\RoutingBundle\Model',
```
dantleech: Can't figure out why we need this? Is it not enough to just ensure the other pass doesn't get triggered?

ahilles107: As in most of cases i will extend base model class (or even CMF Route class) but with custom mapping done in my bundle. If it would be missing then whole mapping for all properties defined in model (mapped superclass) would need to be redeclared in other mapping. Or this config directory would need to be added to `config.yml` mapping.

ahilles107:
Hey, what do You think? It can be merged?

dantleech:
@dbu ?

DependencyInjection/CmfRoutingExtension.php
```diff
@@ -282,6 +282,12 @@ private function loadOrmProvider($config, XmlFileLoader $loader, ContainerBuilde
 
         $container->setParameter('cmf_routing.backend_type_orm', true);
         $container->setParameter('cmf_routing.dynamic.persistence.orm.manager_name', $config['manager_name']);
+        $container->setParameter('cmf_routing.dynamic.persistence.orm.route_class', $config['route_class']);
+        if (false !== strpos($config['route_class'], 'Symfony\Cmf\Bundle\RoutingBundle\Doctrine\Orm')) {
```
dbu: why not an exact string comparison for the ORM route class?

and if we can do that, please use the ::class notation instead of a string for the class name

DependencyInjection/Configuration.php
```diff
@@ -163,6 +163,7 @@ private function addDynamicSection(ArrayNodeDefinition $root)
                                     ->canBeEnabled()
                                     ->children()
                                         ->scalarNode('manager_name')->defaultNull()->end()
+                                        ->scalarNode('route_class')->defaultValue('Symfony\Cmf\Bundle\RoutingBundle\Doctrine\Orm\Route')->end()
```
dbu: please use Route::class and a `use` at the top of the file

ahilles107:
done

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 Changelog
 =========
 
+ * **2016-10-10**: [ORM] Add option to use custom Route class in ORM
```
dbu: please also say that `cmf_routing.backend_type_orm` has been renamed to `cmf_routing.backend_type_orm_default`. somebody might have relied on this to be present, and it will be hard to discover this change when upgrading to 2.0

dbu: you are right. alright then.

ahilles107:
Now only docs are missing ;) https://github.com/symfony-cmf/symfony-cmf-docs/pull/779

dbu:
yep, i waited for the build to finish, to be sure we have no typos or
other reasons for build failure :-) its green now, so i merged.

thanks again for adding this feature!

ahilles107:
Thanks for building whole CMF ;) We are using few bundles in [superdesk/web-publisher](https://github.com/superdesk/web-publisher) project.

dbu:
great to hear. if you want, you can add that project to the list of
projects using the cmf, at
https://github.com/symfony-cmf/symfony-cmf-website/blob/master/src/AppBundle/Resources/data/page.yml#L1018

PR #369, opened 2016-12-02
Move sonata only

ElectricMaxxx:
| Q             | A
| ------------- | ---
| Bug fix?      | [no]
| New feature?  | [no]
| BC breaks?    | [yes]
| Deprecations? | [no]
| Tests pass?   | [yes]
| License       | MIT
| Doc PR        | Todo


dbu:
i just created the styleci cleanup separately and merged that to master. can you please rebase this branch? i would also like to squash the commits before merging. will that be ok or does that create problems with the other branch where you add form types?

ElectricMaxxx:
@dbu as i cherry picked the commits, they are new commits. it would be easy to pick the others too but i think the diff of the closed one to master will show the form types.

ElectricMaxxx:
Rebased and squashed, Ready to review/merge.

ElectricMaxxx:
https://github.com/symfony-cmf/routing-bundle/issues/374 to remove when we have collected the stuff we need in core-bundle.

Resources/config/schema/routing-1.0.xsd
```diff
@@ -83,7 +83,6 @@
         <xsd:attribute name="enabled" type="xsd:boolean" />
         <xsd:attribute name="manager-name" type="xsd:string" />
         <xsd:attribute name="content-basepath" type="xsd:string" />
-        <xsd:attribute name="use-sonata-admin" type="enable_auto" />
```
dbu: @wouterj does that mean we have to rename this file to routing-2.0.xsd?

dbu:
looks good to me apart from the question on the xml schema

Resources/config/schema/routing-1.0.xsd
wouterj: Well, I don't think so. I think we should maybe even remove the version. This is why:

 * We don't host the XSD, so IDE's can never fetch it by URL;
 * Symfony will always validate against the local XSD, which is the latest;
 * Renaming this to 2.0 means everyone has to update their schemaLocations to `routing-2.0.xsd`, even though they didn't use this attributed. Or we have to ship both 1.0 and 2.0, but we don't support both, we only support 2.0;
 * [Symfony core doesn't do versioning either](https://github.com/symfony/symfony/issues/20635).

dbu: makes sense. except if we rename it now people still have to update their IDE. if you want to do it we can, but that should not block this PR then.

PR #372, opened 2016-12-02
avoid validation warmup in symfony 3.2 if no phpcr

### LINKED ISSUE #370: Symfony 3.2 support/bug ?
Hi,

We our the creators of the KunstmaanBundlesCMS for symfony. We are using the symfony-cmf/routing-bundle in our standard edition and we are trying to upgrade now to symfony 3.2. 

But, we have some issues.

When we update to symfony 3.2 we get the following exception:

Interface 'Doctrine\ODM\PHPCR\HierarchyInterface' not found in /home/projects/willfeed/data/willfeed/vendor/symfony-cmf/routing-bundle/Doctrine/Phpcr/Route.php on line 27

That's indeed correct because we do not install the the doctrine-odm package. But in previous symfony versions, this works and we do not get a error there. I think the problem is that as of symfony 3.2 all the classes are automatically warmed up for caching purposes. Therefor symfony will also pass by the Symfony\Cmf\Bundle\RoutingBundle\Doctrine\Phpcr\Route class and will find an interface there that does not exists.

Can you guys take a look at this, or correct me if we are doing something wrong?

Cheers!

ElectricMaxxx:
| Q             | A
| ------------- | ---
| Bug fix?      | [yes]
| Tests pass?   | [yes]
fix #370


wouterj:
This is not how we should test against multiple version. Only add one job using the matrix.include option.

dbu:
@sandergo90 can you please verify if this branch fixes the problem for you?

dbu:
@wouterj would you have an idea what the PARSE_CONSTANT error on travis is about? it happens only for symfony 3.2 and i don't get the error - in symfony 3.2 that constant is in fact defined and phpstorm finds it.

DependencyInjection/Compiler/ValidationPass.php
```diff
@@ -0,0 +1,31 @@
...
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+
+/**
+ * Register validation files if their storage is activated.
+ */
+class ValidationPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container)
+    {
+        // only replace the default router by overwriting the 'router' alias if config tells us to
```
wouterj: this comment seems like a copy/paste error

wouterj:
> @wouterj would you have an idea what the PARSE_CONSTANT error on travis is about? it happens only for symfony 3.2 and i don't get the error - in symfony 3.2 that constant is in fact defined and phpstorm finds it.

Yes I do. It's because PHPunit is built with symfony/yaml 3.0 (which doesn't have this constant). This dep overrides the dep in the vendor directory.

To avoid this, I changed the Travis script to use the Simple PHPunit wrapper from the PHPunitBridge. This builds a phpunit version with the locally installed yaml component.

dbu:
thanks a lot for the help, wouter!

wouterj:
this quite heavily broke everything. Not sure why yet https://travis-ci.org/symfony-cmf/sonata-admin-integration-bundle/jobs/181183400

dbu:
"The mapping file "/home/travis/build/symfony-cmf/sonata-admin-integration-bundle/vendor/symfony-cmf/routing-bundle/DependencyInjection/Compiler../../Resources/config/validation-phpcr.xml" does not exist in . (which is being imported from "/home/travis/build/symfony-cmf/sonata-admin-integration-bundle/tests/Resources/app/config/routing.yml" looks like something is going quite wrong with the routing bundle fix we just merged

dbu:
see #376

PR #375, opened 2016-12-04
adjust to new file structure

### LINKED ISSUE #367: New File Structure

dbu:
fix #367


tests/Unit/DependencyInjection/XmlSchemaTest.php
```diff
@@ -21,7 +21,7 @@ class XmlSchemaTest extends XmlSchemaTestCase
     public function setUp()
     {
         $this->fixturesPath = __DIR__.'/../../Resources/Fixtures/config/';
-        $this->schemaPath = __DIR__.'/../../../Resources/config/schema/routing-1.0.xsd';
+        $this->schemaPath = __DIR__.'/../../../src/Resources/config/schema/routing-1.0.xsd';
```
dbu: as we move this file anyways, should i remove the -1.0 suffix?

wouterj: From an end-user point of view, no change is needed at the moment. But we can maybe make this BC break once in the 2.0 release?

dbu:
the folder of the schema file changes, so people that configured it need to
adjust anyways, i think?

PR #382, opened 2017-01-20
Keep support for PHPCR-ODM 1.4

dbu:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

as discussed on slack, lets keep support for phpcr-odm 1.4 so we can release without needing phpcr-odm 2.0 first.

### DIFF
--- a/src/Doctrine/Phpcr/RedirectRoute.php
+++ b/src/Doctrine/Phpcr/RedirectRoute.php
@@ -100,6 +100,8 @@ class RedirectRoute extends RedirectRouteModel implements PrefixInterface, Hiera
      */
     public function setParent($parent)
     {
+        @trigger_error('The '.__METHOD__.'() method is deprecated and will be removed in version 3.0. Use setParentDocument() instead.', E_USER_DEPRECATED);
+
         return $this->setParentDocument($parent);
     }
 
@@ -109,6 +111,8 @@ class RedirectRoute extends RedirectRouteModel implements PrefixInterface, Hiera
      */
     public function getParent()
     {
+        @trigger_error('The '.__METHOD__.'() method is deprecated and will be removed in version 3.0. Use getParentDocument() instead.', E_USER_DEPRECATED);
+
         return $this->getParentDocument();
     }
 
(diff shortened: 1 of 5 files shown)


src/Doctrine/Phpcr/RedirectRoute.php
```diff
@@ -95,6 +95,24 @@ public function getParentDocument()
     }
 
     /**
+     * @deprecated For BC with the PHPCR-ODM 1.4 HierarchyInterface
+     * @see setParentDocument
+     */
+    public function setParent($parent)
+    {
+        return $this->setParentDocument($parent);
```
wouterj: Please also add deprecation triggers:

```php
@trigger_error('The '.__METHOD__.'() method is deprecated and will be removed in version 3.0. Use setParentDocument() instead.', E_USER_DEPRECATED);
```

dbu: right. just pushed them.

src/Doctrine/Phpcr/Route.php
```diff
@@ -105,6 +105,28 @@ public function getParentDocument()
...
+        return $this->setParentDocument($parent);
+    }
+
+    /**
+     * @deprecated For BC with the PHPCR-ODM 1.4 HierarchyInterface
+     * @see getParentDocument
+     */
+    public function getParent()
+    {
+        @trigger_error('The '.__METHOD__.'() method is deprecated and will be removed in version 3.0. Use setParentDocument() instead.', E_USER_DEPRECATED);
```
wouterj: should be `getParentDocument()` here (and in the other class)

dbu: hah. how to copy paste...

wouterj:
Pushed a fix for the failing build. As Symfony 3.0 is no longer maintained, it has some bugs with latest dependencies, causing our tests to fail. We should simply stop testing against this dep.

PR #388, opened 2017-02-06
do not rename the template in dynamic router

dbu:
$template is in line with the FrameworkBundle TemplateController

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | travis-ci
| Fixed tickets | -
| License       | MIT
| Doc PR        | TODO

together with https://github.com/symfony-cmf/content-bundle/pull/174

### DIFF
--- a/UPGRADE-2.0.md
+++ b/UPGRADE-2.0.md
@@ -21,7 +21,7 @@
                          - Symfony\Cmf\Component\Routing\RouteReferrersReadInterface
    ```
 
-    After:
+   After:
        
    ```yaml
         # app/config/config.yml
@@ -53,7 +53,7 @@
                            - cmf_routing.redirect_route_admin
    ```
 
-    After:
+   After:
        
    ```yaml
         # app/config/config.yml
@@ -70,7 +70,8 @@
 
  * The settings `admin_basepath` and `content_basepath` are only relevant for the admin and thus have been moved as well.
 
-   **Before**
+   Before:
+
    ```yaml
    cmf_routing:
        # ...
@@ -93,7 +94,8 @@
    </config>
    ```
 
-   **After**
+   After:
+
    ```yaml
    cmf_sonata_phpcr_admin_integration:
        # ...
@@ -113,14 +115,15 @@
    </config>
    ```
 
-## Route Model
+### Route Model
 
  * Removed `getAddFormatPattern()`/`setAddFormatPattern()` from the model
    `Route` and `getAddTrailingSlash()`/`setAddTrailingSlash()` from the PHPCR
    `Route`. Use `getOption()`/`setOption()` with `'add_format_pattern'` or
    `'add_trailing_slash'` instead.
 
-   **Before**
+   Before:
+
    ```php
    $route->setAddFormatPattern(true);
    $route->setAddTrailingSlash(true);
@@ -130,7 +133,8 @@
    }
    ```
 
-   **After**
+   After:
+
    ```php
    $route->setOption('add_format_pattern', true);
    $route->setOption('add_trailing_slash', true);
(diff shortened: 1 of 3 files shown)


ElectricMaxxx:
Doing so all custom controllers, should call its template parameter `$template`, right? I thought we have some kind of configuration here to be flexible.

dbu:
they have to, if they want parameter automatch. but i think its worth it to avoid the collision with FrameworkBundle. we don't have configuration for that name, only a constant - but a parameter variable name can't use a constant afaik. however, you could write your controller expecting the Request instead, and then get the attribute with the constant...

wouterj:
A couple of remarks here:

 * Why at all do this renaming of `_template` to `contentTemplate`?
 * How can this conflict with the framework's template controller? It'll always require `template`, not `_template`.
 * Why remove `_template` when renaming it to `contentTemplate`? Does it hurt having a duplicated request attribute?
 * If we want to do this change, we should release a new 1.4 version that triggers a deprecation warning when using `contentTemplate`. It should also come with 2 attributes: `template` and `contentTemplate` to allow forward compatibility.

UPGRADE-2.0.md
```diff
@@ -198,3 +198,37 @@
...
+   ```php
+   public function documentAction($contentDocument, $contentTemplate) {
+       ...
+   ```
+
+   **After**
+   ```php
+   public function documentAction($contentDocument, $template) {
+       ...
+   ```
```
wouterj: this file doesn't seem right. The lines 208 - 222 should be removed, right?

dbu:
>   * Why at all do this renaming of |_template| to |contentTemplate|?

having a request attribute called _template triggers an exception by the
framework bundle that _template is reserved for the `@Template` annotation.

>   * How can this conflict with the framework's template controller?
>     It'll always require |template|, not |_template|.

i feel it makes sense to use the same name. its the same concept.

>   * Why remove |_template| when renaming it to |contentTemplate|? Does
>     it hurt having a duplicated request attribute?

yes, having _template triggers an exception.

>   * If we want to do this change, we should release a new 1.4 version
>     that triggers a deprecation warning when using |contentTemplate|. It
>     should also come with 2 attributes: |template| and |contentTemplate|
>     to allow forward compatibility.

putting both into the request attributes seems like a good idea. i don't
know how we could do a deprecation warning however - something in the
core of symfony looks at controller method parameters and tries to match
them with request attribute names.

wouterj:
Let's update this PR to use both `contentTemplate` and `template` in 2.0. We then have more time to see how we can trigger deprecations for the `contentTemplate` (and I believe we should propose to stop throwing an exception when `_template` is present, it doesn't hurt anybody).

dbu:
yeah, i was going to say thats not our fault :-)

UPGRADE-2.0.md
dbu: indeed, messed up with copy-pasting it seems.

dbu:
i think this is now ready

src/Routing/DynamicRouter.php
```diff
@@ -41,7 +41,15 @@ class DynamicRouter extends BaseDynamicRouter
...
-    const CONTENT_TEMPLATE = 'contentTemplate';
+    const CONTENT_TEMPLATE = 'template';
+
+    /**
+     * RoutingBundle 1.* put the template into the request attributes as "contentTemplate".
+     *
+     * This name is deprecated in favor of "template" for consistency with
+     * FrameworkBundle and will be removed in RoutingBundle 3.
+     */
+    const BC_CONTENT_TEMPLATE = 'contentTemplate';
```
wouterj: I don't think we have to use a constant for the BC version, it'll only cause more conflicts (people shouldn't start using this constant)

wouterj:
Applied my comment and slightly tweaked the changelog and upgrade guide. Feel free to merge if you agree with my changes.

dbu:
good point, thanks for the cleanup and the merge!

PR #395, opened 2017-09-23
add dependencies for a clean install

ElectricMaxxx:
I try to get a valid dependency situation, for installing routing bundle on its own with symfony >= 3.3

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,10 @@
     "require": {
         "php": "^5.6|^7.0",
         "symfony-cmf/routing": "^2.0",
-        "symfony/framework-bundle": "^2.8|^3.0"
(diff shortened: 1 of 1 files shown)


ElectricMaxxx:
Having https://github.com/symfony/recipes-contrib/pull/91#issuecomment-331666757 in mind sounds like we have to many dependencies for doing nothing, means for no configuration. It also means that i diddn't got it working without:
```
framework:
    templating: { engines: ['twig'] }
````
the chain of dependencies is:
validator -> templating -> framework configuration -> assets -> twig-bundle

ElectricMaxxx:
We should not add it we should replace it.

PR #397, opened 2017-09-25
use twig loader to validate template's existence

### LINKED ISSUE #396: Remove symfony/templating dependency
That component will be deprecated, so we should replace it to stay symfony 4.0 compatible.

ElectricMaxxx:
| Q             | A
| ------------- | ---
| Bug fix?      | [no]
| New feature?  | [no]
| BC breaks?    | [no]
| Deprecations? | [no]
| Tests pass?   | [yes]
| Fixed tickets | [#396]
| License       | MIT

fix #396 

I  replaced templating by a twig loader based on the comment of @fabpot (maybe you can give us a internal in review)
https://github.com/symfony/recipes-contrib/pull/91#issuecomment-331666757

@dbu @wouterj I have no clue if we can release that on a patch release or if it is an BC-Break. From my POV we didn't change any external API, so a patch release would be ok.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+2.0.1
+-----
+
+* **2017-09-25**: This bundle can now also directly use the Twig loader instead of the deprecated templating
+  component. Symfony FrameworkBundle no longer requires symfony/templating since 3.2. If the templating component
+  is available in your application, it is however still used for BC.
+
 2.0.0
 -----
 
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,8 @@
     "require": {
         "php": "^5.6|^7.0",
         "symfony-cmf/routing": "^2.0",
-        "symfony/framework-bundle": "^2.8|^3.0"
+        "symfony/framework-bundle": "^2.8|^3.0",
+        "symfony/twig-bundle": "^2.8|^3.0"
     },
     "require-dev": {
         "symfony-cmf/testing": "^2.0",
--- a/src/DependencyInjection/Compiler/TemplatingValidatorPass.php
+++ b/src/DependencyInjection/Compiler/TemplatingValidatorPass.php
@@ -1,0 +1,39 @@
+<?php
+
+/*
+ * This file is part of the Symfony CMF package.
+ *
+ * (c) 2011-2017 Symfony CMF
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Cmf\Bundle\RoutingBundle\DependencyInjection\Compiler;
+
+use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface;
+use Symfony\Cmf\Bundle\RoutingBundle\Validator\Constraints\RouteDefaultsTemplatingValidator;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * To avoid a BC-Break: If templating component exists, we will use the validator using general templating
+ * from FrameworkBundle.
+ *
+ * @author Maximilian Berghoff <Maximilian.Berghoff@mayflower.de>
+ */
+class TemplatingValidatorPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        if (interface_exists(EngineInterface::class) && $container->has('templating')) {
+            $templatingDefinition = $container->findDefinition('templating');
+            $validatorDefinition = $container->getDefinition('cmf_routing.validator.route_defaults');
+            $validatorDefinition->setClass(RouteDefaultsTemplatingValidator::class);
+            $validatorDefinition->replaceArgument(1, $templatingDefinition);
+        }
+    }
+}
--- a/src/Resources/config/validators.xml
+++ b/src/Resources/config/validators.xml
@@ -5,9 +5,12 @@
            xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
 
     <services>
-        <service id="cmf_routing.validator.route_defaults" class="Symfony\Cmf\Bundle\RoutingBundle\Validator\Constraints\RouteDefaultsValidator">
+        <service
+                id="cmf_routing.validator.route_defaults"
+                class="Symfony\Cmf\Bundle\RoutingBundle\Validator\Constraints\RouteDefaultsTwigValidator"
+        >
             <argument id="controller_resolver" type="service"/>
-            <argument id="templating" type="service"/>
+            <argument id="twig.loader" type="service"/>
             <tag name="validator.constraint_validator" alias="cmf_routing.validator.route_defaults" />
         </service>
     </services>
--- a/src/Validator/Constraints/RouteDefaultsTemplatingValidator.php
+++ b/src/Validator/Constraints/RouteDefaultsTemplatingValidator.php
@@ -1,0 +1,62 @@
+<?php
+
+/*
+ * This file is part of the Symfony CMF package.
+ *
+ * (c) 2011-2017 Symfony CMF
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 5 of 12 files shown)


ElectricMaxxx:
The build fail is based on a parameter `kernel.project_dir`, which exists on Symfony 3.3 and not before. I think we can set the `twig.path` on whatever we whant in testing, in an app the user should set it to a value of its needs.

composer.json
```diff
@@ -14,7 +14,8 @@
     "require": {
         "php": "^5.6|^7.0",
         "symfony-cmf/routing": "^2.0",
-        "symfony/framework-bundle": "^2.8|^3.0"
+        "symfony/framework-bundle": "^2.8|^3.0",
+        "symfony/twig-bundle": "^2.8|^3.0"
```
dbu: are you sure we need the twig bundle? would twig/twig not be enough?

src/Validator/Constraints/RouteDefaultsValidator.php
```diff
@@ -11,21 +11,25 @@
...
     private $controllerResolver;
-    private $templating;
 
-    public function __construct(ControllerResolverInterface $controllerResolver, EngineInterface $templating)
+    /**
+     * @var LoaderInterface
+     */
+    private $twig;
+
+    public function __construct(ControllerResolverInterface $controllerResolver, LoaderInterface $twig)
```
dbu: if we would omit the typehint here, and validate the class in the constructor to be either loader or engine, we could be BC. not sure what the best practice is for that.

src/Resources/config/validators.xml
```diff
@@ -7,7 +7,7 @@
     <services>
         <service id="cmf_routing.validator.route_defaults" class="Symfony\Cmf\Bundle\RoutingBundle\Validator\Constraints\RouteDefaultsValidator">
             <argument id="controller_resolver" type="service"/>
-            <argument id="templating" type="service"/>
+            <argument id="twig.loader" type="service"/>
```
dbu: i guess the real BC break is this. what if people use a non-twig template? then things will suddenly stop working. could we find a way to use templating if it exists, but twig.loader otherwise? maybe with an expression language check?

composer.json
dbu: ah sorry, we need the bundle because we want the twig.loader service. this change is fine imho.

src/Resources/config/validators.xml
ElectricMaxxx: I would suggest to do it like here:
https://github.com/symfony/framework-bundle/blob/master/Controller/ControllerTrait.php#L211
in a kind of compiler pass. We would have two validator classes then. If the user does a `composer require symfony/templating` a validator using templating would overwrite the default validator using twig in our container.

ElectricMaxxx: But we have to document that `symfony/templating` should be added manually from Symfony 3.2 on, cause it was removed from frameworkbundle.

dbu: +1 for the compiler pass to update the service definition for the validator to use the templating service if available. i would just remove the typehint and say it must be one or the other class.

we never required the templating here, so explaining how to update is "out of scope" and part of upgrading symfony itself. we can just explain what happened in the changelog, like " RoutingBundle can now also directly work with Twig. Symfony FrameworkBundle no longer requires symfony/templating since 3.2 and the component is deprecated. If the templating component is available in your application, it is however still used for BC."

ElectricMaxxx: I would avoid to remove the type hint, cause the fact that both services got the same method is random.  So yes, we should have two services, which can be changed by a compiler pass.

to the templating problem as a dependency:
When having a Symfony App with version >= 3.2 and doing a simple `composer require symfony-cmf/routing-bundle` to add routing functions, you will get an error that templating service isn't available. The reason is: templating was removed from framework-bundle. So have to do a `composer require symfony/templating` before for those apps. So requiring templating is not a question of which templating engine to chose, it is a question of the symfony version.

ElectricMaxxx:
I think we will get the same trouble in content-bundle, where our Controller uses the engine.

src/DependencyInjection/Compiler/TemplatingValidatorPass.php
```diff
@@ -0,0 +1,29 @@
...
+ * @author Maximilian Berghoff <Maximilian.Berghoff@mayflower.de>
+ */
+class TemplatingValidatorPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        if (class_exists(EngineInterface::class) && $container->has('templating')) {
```
ElectricMaxxx: `templating` could exists in FrameworkBundle (before Symfony 3.2) and in a single namespace (3.2 and later). The one in FrameworkBundle is the one we need to use, as it is the one we used before. The one in the new component looks same but got an other namespace and would break the contructor then. So i will cover with this check the former `templating` to avoid a BC-Break. From Symfony 3.2 the user should use TwigBundle. But i have no clue how to test that one except from checking it out in sandbox with different symfony versions, any other idea?

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 Changelog
 =========
 
+* **2017-02-17**: Remove usage of `templating` component to validate template existence,
+  use twig loader instead.
```
dbu: can we change this to "This bundle can now also directly use the Twig loader instead of the deprecated templating component. Symfony FrameworkBundle no longer requires symfony/templating since 3.2. If the templating component is available in your application, it is however still used for BC."

and please check the date, i'd guess 2017-09-25 would be correct

ElectricMaxxx: I should also add the path to the bundlle ;-)

When removing it later, we should deprecate it.

src/Validator/Constraints/RouteDefaultsTemplatingValidator.php
```diff
@@ -17,9 +17,18 @@
 use Symfony\Component\Validator\Constraint;
 use Symfony\Component\Validator\ConstraintValidator;
 
-class RouteDefaultsValidator extends ConstraintValidator
+/**
+ * @deprecated Will be removed in 3.0. Use RouteDefaultsTwigValidator with twig as template engine instead.
```
ElectricMaxxx: @dbu is that notice got enough?

src/DependencyInjection/Compiler/TemplatingValidatorPass.php
```diff
@@ -0,0 +1,39 @@
...
+            $templatingDefinition = $container->getDefinition('templating');
+            $validatorDefinition = $container->getDefinition('cmf_routing.validator.route_defaults');
+            $validatorDefinition->setClass(RouteDefaultsTemplatingValidator::class);
+            $validatorDefinition->setArgument('$templating', $templatingDefinition);
```
dbu: does this syntax work with all symfony versions or was that added in 3.2 or so?

ElectricMaxxx: you can be right, That syntax is new, so i need the index of the argument. Is the counting 0-based?

dbu: yes, its 0 based. can you check in the sandbox or so if it actually works? i think building a test is a lot of hassle, i am ok if we know it worked at the time we add it...

ElectricMaxxx:
uhh, this was a hard thing at the end, but now it is green and it works.

src/DependencyInjection/Compiler/TemplatingValidatorPass.php
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        if (interface_exists(EngineInterface::class) && $container->has('templating')) {
+            $templatingDefinition = $container->findDefinition('templating');
+            $validatorDefinition = $container->getDefinition('cmf_routing.validator.route_defaults');
+            $validatorDefinition->setClass(RouteDefaultsTemplatingValidator::class);
+            $validatorDefinition->replaceArgument(1, $templatingDefinition);
```
stof: this is wrong. You must not copy the definition in the argument. This would create a separate instance of the service. You must use `new Reference('templating')` as value of the argument, to reference the existing service

ElectricMaxxx: @stof  but i would then inject it by `replaceArgument()` right? but thanks, had some trouble how to use aliases with service definitions.

PR #400, opened 2017-09-26
remove interface check at all

ElectricMaxxx:
and the next issue

### DIFF
--- a/src/DependencyInjection/Compiler/TemplatingValidatorPass.php
+++ b/src/DependencyInjection/Compiler/TemplatingValidatorPass.php
@@ -28,7 +28,7 @@ class TemplatingValidatorPass implements CompilerPassInterface
      */
     public function process(ContainerBuilder $container)
     {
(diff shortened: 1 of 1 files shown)


dbu:
so the `templating` service always is of the right interface if it exists, after all?

ElectricMaxxx:
Yes, as long as we do not `composer require symfony/templating`, which is kind of derpecated and configure it in framework configuration. the last is done on older apps, but not with the new component. 

I really had trouble yesterday to get it working. as `interface_exists()` seems to require the interface and sends an warning. 

you can now create a symfony skeleton app `composer create-project "symfony/skeleton:^3.3"`, require cli (the comandline and cache calls won't work) `composer req cli` and call our reciepe then: `SYMFONY_ENDPOINT=https://symfony.sh/r/github.com/symfony/recipes-contrib/91 composer remove symfony-cmf/routing-bundle` and it should go green. In sandbox we have the former configuration with the old framework-bundle and that works fine too.

dbu:
great job, thanks a lot max!

PR #429, opened 2018-12-10
start implementing ORM Route

ElectricMaxxx:
I'd like to hadd some ORM functionoal base tests to introduce cherries?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,8 @@
         "php": "^7.1",
         "symfony-cmf/routing": "^2.1.0",
         "symfony/framework-bundle": "^2.8 || ^3.3 || ^4.0",
-        "twig/twig": "^1.35 || ^2.4.4"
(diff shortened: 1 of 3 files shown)


ElectricMaxxx:
Ahhh .. now it becomes clearer: At the end we do not have a `path` property mapped to our ORM model. It is a setter/getter living in Symfony route model.

So we would use `staticPrefix` at all?

Shouldn't we use a path also? for easier querying? or not?

ElectricMaxxx:
Now it becomes clearer. When asking for `staticPrefix` as done in `RouteProvider`, which i do not like to use in that functional test, i get my routes. But we do have a quite different behavior then in PHPCR world. 

- route content is not managed at all
- variable pattern with our without `/` at the beginning
- some listeners, which do not fetch falsy prefixes here

ElectricMaxxx:
@dbu you found some time to have a look in this one here?

PR #434, opened 2019-01-29
DevKit updates

ElectricMaxxx:
Let's start it here also

ToDo:
* [x] fix tests
* [x] increase phpunit version


.travis.yml
```diff
@@ -26,23 +26,23 @@ cache:
...
 
 env:
-  matrix: SYMFONY_VERSION=4.0.*
+  matrix: SYMFONY_VERSION=4.2.*
   global:
-    - SYMFONY_DEPRECATIONS_HELPER="/.*each.*/"
+    - SYMFONY_DEPRECATIONS_HELPER="weak"
     - SYMFONY_PHPUNIT_DIR=.phpunit SYMFONY_PHPUNIT_REMOVE="symfony/yaml"
-    - SYMFONY_PHPUNIT_VERSION=6
-    - TEST_INSTALLATION=false
```
dbu: is this now the default? we should only test the installation in one build, not all of them

.travis.yml
```diff
@@ -26,23 +26,23 @@ cache:
...
+    - KERNEL_CLASS=Symfony\Cmf\Bundle\RoutingBundle\Tests\Fixtures\App\Kernel
+    - SYMFONY_PHPUNIT_VERSION=5.7.26
```
dbu: please do not roll back to phpunit 5, i upgraded this bundle to 6 yesterday

.travis.yml
```diff
@@ -26,23 +26,23 @@ cache:
...
 env:
-  matrix: SYMFONY_VERSION=4.0.*
+  matrix: SYMFONY_VERSION=4.2.*
   global:
-    - SYMFONY_DEPRECATIONS_HELPER="/.*each.*/"
+    - SYMFONY_DEPRECATIONS_HELPER="weak"
     - SYMFONY_PHPUNIT_DIR=.phpunit SYMFONY_PHPUNIT_REMOVE="symfony/yaml"
-    - SYMFONY_PHPUNIT_VERSION=6
-    - TEST_INSTALLATION=false
+    - KERNEL_CLASS=Symfony\Cmf\Bundle\RoutingBundle\Tests\Fixtures\App\Kernel
```
dbu: this is no longer needed and does nothing. it is defined in phpunit.xml.dist

.travis.yml
```diff
@@ -26,23 +26,23 @@ cache:
...
-    - php: 7.1
-      env: COMPOSER_FLAGS="--prefer-lowest" SYMFONY_VERSION=2.8.* SYMFONY_DEPRECATIONS_HELPER=weak
-    - php: 7.2
-      env: SYMFONY_VERSION=3.3.*
-    - php: 7.2
-      env: SYMFONY_VERSION=3.4.*
+    - php: "7.3"
+      env: STABILITY="dev" SYMFONY_VERSION="4.2.*"
+    - php: "7.2"
+      env: COMPOSER_FLAGS="--prefer-lowest" SYMFONY_VERSION="3.4.*" SYMFONY_DEPRECATIONS_HELPER="weak"
```
dbu: are you dropping symfony 2.8 support? i think it makes sense, 2.8 is really ancient. only a bit surprised to have that as part of a general devkit update.... we might bump the minor version for it and add a note in the changelog though.

.travis.yml
```diff
@@ -26,23 +26,23 @@ cache:
...
-    - php: 7.2
-      env: SYMFONY_VERSION=4.0.*
-    - php: 7.1
-      env: COMPOSER_FLAGS="--prefer-lowest" SYMFONY_VERSION=2.8.* SYMFONY_DEPRECATIONS_HELPER=weak
-    - php: 7.2
-      env: SYMFONY_VERSION=3.3.*
-    - php: 7.2
-      env: SYMFONY_VERSION=3.4.*
+    - php: "7.3"
+      env: STABILITY="dev" SYMFONY_VERSION="4.2.*"
```
dbu: why build everything with stability dev? one build with dev is fine for me, but the rest should be stable

README.md
```diff
@@ -8,12 +8,13 @@
...
 [![Daily Downloads](https://poser.pugx.org/symfony-cmf/routing-bundle/d/daily)](https://packagist.org/packages/symfony-cmf/routing-bundle)
 
-Version/Branch | Travis | Coveralls |
------- | ------ | --------- |
-2.1.1   | [![Build Status][travis_stable_badge]][travis_link]     | [![Coverage Status][coveralls_stable_badge]][coveralls_stable_link]     |
-master | [![Build Status][travis_unstable_badge]][travis_link] | [![Coverage Status][coveralls_unstable_badge]][coveralls_unstable_link] |
+Branch | Travis | Coveralls | Scrutinizer |
+------ | ------ | --------- | ----------- |
+2.1   | [![Build Status][travis_stable_badge]][travis_stable_link]     | [![Coverage Status][coveralls_stable_badge]][coveralls_stable_link]     | [![Scrutinizer Status][scrutinizer_stable_badge]][scrutinizer_stable_link] |
+3.0-dev | [![Build Status][travis_unstable_badge]][travis_unstable_link] | [![Coverage Status][coveralls_unstable_badge]][coveralls_unstable_link] | [![Scrutinizer Status][scrutinizer_unstable_badge]][scrutinizer_unstable_link] |
```
dbu: there is no branch 3.0-dev, and master represents 2.2. i'd keep master here and never point to upcoming not-yet-master branches as the dev version.

composer.json
```diff
@@ -15,20 +15,20 @@
         }
     ],
     "require": {
-        "php": "^7.1",
+        "php": "^7.2",
```
dbu: why drop 7.1 support?

composer.json
```diff
@@ -15,20 +15,20 @@
...
         "symfony-cmf/routing": "^2.1.0",
-        "symfony/framework-bundle": "^2.8 || ^3.3 || ^4.0",
+        "symfony/framework-bundle": "^3.4 || ^4.0",
         "twig/twig": "^1.35 || ^2.4.4"
     },
     "require-dev": {
         "doctrine/phpcr-odm": "^1.4|^2.0",
-        "symfony/phpunit-bridge": "^4.2.2",
+        "symfony/phpunit-bridge": "^3.4 || ^4.0",
```
dbu: i locked onto 4.2.2 for the lowest build support. and as this is a dev dependency and does not depend on other symfony things, there is no value in allowing flexibility. we should always go with the latest version.

.travis.yml
ElectricMaxxx: no, i created 2.1 branch for latest stable, master goes towards 3.0-dev now.

ElectricMaxxx: and yes, we do have the controll over symfony and php version out of our dev-kit

README.md
ElectricMaxxx: when requiring `symfony-cmf/routing-bundle:3.0-dev` we should get master now, when this PR is merged, isn't it?

ElectricMaxxx:
Btw: no problems to start with a single repository and then roll it out to the other ones, but we are tightly coupled to our testing component in dev. So we have to do them both hand in hand.

composer.json
ElectricMaxxx: :+1: i currently realized that my local ubuntu works on 7.1 only :-)

.travis.yml
ElectricMaxxx: i thought, we build with dev when having the a point where we are on dev in master? So, if we require a dev version for our testing component i.e.?

.travis.yml
```diff
@@ -16,7 +16,7 @@
 language: php
 
 php:
-  - 7.1
+  - "7.1"
```
dbu: are you sure about this? i have never seen it with quotes

.travis.yml
dbu: i think the intention was that we would switch it to fail the build once it was reliable enough. in the beginning we did it for supporting symfony 4 and it failed everywhere because we did not support it yet.

.travis.yml
dbu: no, we want one build with dev to see if future symfony versions etc will pose some regression. but otherwise stable dependencies make more sense, they are what our users will have.

additionally, using dev can also hide problems when no stable release of a dependency is available.

i'd prefer to leave this as it was.

.travis.yml
ElectricMaxxx: My IDE complains about non strings in YAML, what can go wrong? The versions in the matrix are doing their job.

.travis.yml
dbu: why do you want to start a version 3, are there big BC breaking changes coming up?

if you have them, it makes sense - though i have to say upfront that i won't have much time to invest into another large-scale effort into the cmf. i would have left things and reviewed contributions and do the most important things like the effort i did over this weekend to make the tests run again.

README.md
dbu: you are not changing the branch-alias in composer.json so there will not be a 3.0-dev version.

dbu: and i would keep this at `master` so its whatever master is at any time. one less thing that we can forget to update :-)

.travis.yml
dbu: i removed this line in [#433](https://github.com/symfony-cmf/routing-bundle/pull/433/files#diff-354f30a63fb0907d4ad57269548329e3L33) and the build was green, so i am sure that we can take it away. it should be defined in phpunit.xml.dist, because then it will also work when running the tests locally.

having it here won't break anything, it is simply not seen and can create confusion if somebody needs to change the value and nothing happens when changing it in this file.

.travis.yml
dbu: right, this is a yaml file :man_facepalming: 
agreed. i know some people who have choice words to say about yaml :-D 
its often a bit sloppy, i remember symfony having changed their parser to become more strict at some point because it did not complain about totally stupid things...
lets keep the quotes.

.travis.yml
ElectricMaxxx: The var is also in the make file, will remove it there also and we will see :-)

dbu: thats many places :-/
locally i was not using the makefile but `vendor/bin/phpunit` directly, which worked fine.

.travis.yml
ElectricMaxxx: no worries about the time. As long as i got those initial and general hints as here i am happy.
To the BC-Breaks:
- remove SF 2.8 Support
- Use PHPUnit >=6
- SonataIntegration has to switch on FosCkEditor to get SF4 support at all -> changes in naming of our config

At the end there are no real changes in our API, but we drop some dependencies. That was the reason to move into new major. The only thinkg that is growing atm from outer persons is ORM support.

.travis.yml
ElectricMaxxx: oh if that works it would be awesome :-)

.travis.yml
dbu: dropping support for old versions is no BC break. composer.json will prevent from installing an impossible combination. routing does not provide phpunit related things to the outside. and the editor is only in the sonata integration.

i would prefer to do a new minor version for the phpunit and change of supported symfony versions. we could do a new major for testing and sonata integration only. that would make more sense to me. and sonata integration is anyway on a different version number than the rest, and testing used only internally.

ElectricMaxxx: ok, that makes things easier and faster

.travis.yml
ElectricMaxxx: i put it back.

.php_cs
```diff
@@ -0,0 +1,3 @@
+<?php
+// TO BE REMOVED.
```
dbu: yep, the non-dist file should not be under version control. i think it should even be ignored with .gitignore

composer.json
```diff
@@ -17,18 +17,18 @@
     "require": {
         "php": "^7.1",
         "symfony-cmf/routing": "^2.1.0",
-        "symfony/framework-bundle": "^2.8 || ^3.3 || ^4.0",
+        "symfony/framework-bundle": "^3.4 || ^4.0",
```
dbu: i'd do `^3.4 || ^4.1` as sf 4.0 is also EOL

src/Model/Route.php
```diff
@@ -236,7 +226,7 @@ public function getPath()
      * When using PHPCR-ODM, make sure to persist the route before calling this
      * to have the id field initialized.
      */
-    public function setPath($pattern)
+    public function setPath(string $pattern)
```
ElectricMaxxx: @dbu i had to do that change here to avoid a `null === $pattern` check as `strpos` seems not to like `null` anymore. But this change i also have to make in base interface, wdyt?

ElectricMaxxx: ahh, this is SF component :-) so we need to change our default values going into the SF component part.
The major issue is: `id` and `idPrefix` is `null` and goes later into an `strpos($id, $idPrefix)`

dbu: please do a `null === ` check instead of changing the signature. changing that interface is a BC break and its not some kind of internal interface that we expect nobody to have implemented.

for the next major version, its the right thing to make it strict typed (with a changelog note about it), but until then lets not break it.

ElectricMaxxx:
@dbu can you have a look at the tests, please? Do we expose an issue by using more strict code or did i break something through my changes?

ElectricMaxxx:
@dbu nothing to do, my fault.

ElectricMaxxx:
@dbu it took some time but now it is final and green. Any hints to add on changelog? any other issues?

CHANGELOG.md
```diff
@@ -4,7 +4,10 @@ Changelog
 2.2.0(ureleased)
 ---------------
 
-* **YYYY-MM-DD**:
+* **2015-02-26**:introduce PHPCS for style checks again
+* **2015-02-26**: Symfony 4.2 combustibility verified through tests, Symfony 4.3 allowed to break but on watch in tests
+* **2015-02-26**: drop Symfony 2.8 support
+* **2015-02-26**: make use of PHPUnit 6
```
dbu: we are in 2019 ;-) all of the changelog entries need to be fixed.

CHANGELOG.md
```diff
@@ -4,7 +4,10 @@ Changelog
 2.2.0(ureleased)
 ---------------
 
-* **YYYY-MM-DD**:
+* **2015-02-26**:introduce PHPCS for style checks again
+* **2015-02-26**: Symfony 4.2 combustibility verified through tests, Symfony 4.3 allowed to break but on watch in tests
```
dbu: what do you mean by combustibility? i would say "Symfony 4.2 support without warnings." and remove the rest, the changelog is not for us but for users of the bundle.

CHANGELOG.md
```diff
@@ -4,7 +4,10 @@ Changelog
 2.2.0(ureleased)
 ---------------
 
-* **YYYY-MM-DD**:
+* **2015-02-26**:introduce PHPCS for style checks again
```
dbu: i'd remove this line

PR #441, opened 2019-09-17
fixes composer memory limit

killerwolf:
| Q             | A
| ------------- | ---
| Branch?       | "master" for new features / the branch of the current release for fixes
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 
| License       | MIT
| Doc PR        |

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ env:
     - SYMFONY_PHPUNIT_DIR=.phpunit SYMFONY_PHPUNIT_REMOVE="symfony/yaml"
     - SYMFONY_PHPUNIT_VERSION=6
     - TEST_INSTALLATION=false
+    - COMPOSER_MEMORY_LIMIT=-1
 
 matrix:
   include:


dbu:
thanks! it seems that we end up with a "too new" phpunit. we should probably rather change the tests to use the namespaced phpunit classes than pin phpunit to an old version... would you be up to do that as well?

killerwolf:
Crap....symfony-cmf/testing latest relase is so old,

killerwolf:
@dbu  tests of the current code isn't the problem (which rely on the new phpunit namespaced classes) but symfony-cmf/testing.

dbu:
we explicitly request phpunit 6 in https://github.com/symfony-cmf/routing-bundle/blob/master/.travis.yml#L33 - so this looks like a "how could this ever have worked" case. can we require a newer version of the symfony-cmf/testing library?

dbu:
i am fixing this in #443 now. thanks for the effort @killerwolf

PR #442, opened 2019-12-02
Add service alias for the ChainRouterInterface

JoppeDC:
Alias `Symfony\Cmf\Component\Routing\ChainRouterInterface` to `cmf_routing.router` to allow autowiring in strict mode.
Not sure if this is the correct target branch, please check.

| Q             | A
| ------------- | ---
| Branch?       | 2.1? Not sure.
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT
| Doc PR        | reference to the documentation PR, if any

### DIFF
--- a/src/Resources/config/routing-chain.xml
+++ b/src/Resources/config/routing-chain.xml
@@ -11,5 +11,6 @@
                 <argument type="service" id="router.request_context" />
             </call>
         </service>
+        <service id="Symfony\Cmf\Component\Routing\ChainRouterInterface" alias='cmf_routing.router'/>
     </services>
 </container>


dbu:
thanks for the merge request! i fixed the build errors in the master branch - can you please rebase so that we see if things are indeed fixed or if something breaks with this?

dbu:
rebased and works fine. thanks for the contribution!

note that the default mode of using the bundle would be to have the chain router replace the default `router` service. i would hope that autowiring RouterInterface would then give you the chainrouter too. this alias would only be needed if you do something custom where you explicitly care about getting the chain router.

PR #447, opened 2019-12-07
build with symfony 5

dbu:
blocked by https://github.com/symfony-cmf/Routing/pull/240

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ phpunit.xml
 composer.lock
 vendor
 tests/Fixtures/App/var/
+.phpunit.result.cache 
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,8 @@ matrix:
       env: SYMFONY_REQUIRE=4.3.*
     - php: 7.4
       env: SYMFONY_REQUIRE=4.4.*
+    - php: 7.4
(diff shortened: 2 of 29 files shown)


acrobat:
I think these dependencies should be updated to this version constraint, the `phpcr-odm` should be ok I think as 1.5 supports sf5

```
"matthiasnoback/symfony-dependency-injection-test": "^4.1.0",
"matthiasnoback/symfony-config-test": "^4.1.0",
```

dbu:
hm, that did not help. i wonder if it is some change with symfony-cmf/testing that we no longer automatically get a phpcr-implementation and have to require that ourselves?

dbu:
baby steps. now there are further missing dependencies for symfony 4, and for symfony 5 we need to adjust to BC breaks in the noback tests. i have to focus on work now again, if someone has time to try to get it to build, that would be great. 

@alexander-schranz i would prefer to see this building correctly before i tag symfony-cmf/Routing. and from what i understand, you use the bundle as well, so you need this merged anyways before sulu can support symfony 5?

acrobat:
@dbu I will try to jump in and make some progress!

dbu:
wrapped up in #448 , thanks a lot!

dbu:
or rather: we merged #448 to here, need to merge this now :)

PR #449, opened 2020-04-28
use released routing component

dbu:


composer.json
```diff
@@ -27,7 +27,7 @@
         "matthiasnoback/symfony-dependency-injection-test": "^4.1.0",
         "matthiasnoback/symfony-config-test": "^4.1.0",
         "doctrine/orm": "^2.5",
-        "symfony-cmf/testing": "dev-master",
+        "symfony-cmf/testing": "^2.3.0",
```
alexander-schranz, line 30:
```suggestion
        "symfony-cmf/testing": "^3.2",
```

Looks like you accidently changed the false package, think we need to tag symfony-cmf/testing also. And then change `"symfony-cmf/routing": "^2.3",` above

dbu: oops. thanks. i tagged testing and changed both.

PR #467, opened 2021-08-26
Fix create unique index for Oracle DB

FaritSlv:
| Q             | A
| ------------- | ---
| Branch?       | 2.5.1
| Bug fix?      | no
| New feature?  |no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 
| License       | MIT
| Doc PR        |

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "symfony-cmf/routing-bundle",
+    "name": "farit-slv/routing-bundle",
     "type": "symfony-bundle",
     "description": "Symfony RoutingBundle",
     "keywords": [
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "farit-slv/routing-bundle",
+    "name": "symfony-cmf/routing-bundle",
     "type": "symfony-bundle",
     "description": "Symfony RoutingBundle",
     "keywords": [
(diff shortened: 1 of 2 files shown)


dbu (commented):
thank you for the pull request.

the little intricacies of databases. i hope this change does not make regressions for other databases. do i understand correctly that unique-constraints implies that an index is created for the column?

can you please rebase on master? i fixed the github actions build so that we can see the build again.

FaritSlv:
@dbu hello.

This fix creates the correct index for any ORM database.

The point is that there is no such problem in MySQL (Maria DB), that is, it creates a unique index on a column and separately (alias) an index on the same column. In OracleDB it is impossible to create 2 different indexes on the same column. An error will occur that the index has already been created.

MySQL:
image

Correct create index:
image

dbu:
ah, so we even ended up with a duplicated index in mysql? then that fix benefits even those using mysql or postgres.

can you please rebase the branch so that we can see the CI working?

and can you please add a note in the CHANGELOG.md file that quickly explains what changed and why? (it might be that people upgrading will see a doctrine migration for example, to cleanup the duplicated index)

dbu:
that is always the last resort. i will be travelling in the next days, will look into it end of this week.

FaritSlv:
@dbu ready [PR](https://github.com/symfony-cmf/routing-bundle/pull/469)

PR #471, opened 2021-12-30
Symfony 6.0 compatibility

firstred:
| Q             | A
| ------------- | ---
| Branch?       | "master"
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets |
| License       | MIT
| Doc PR        |

This PR makes the bundle compatible with Symfony 6.0.

I ran the [patch-type-declarations](https://wouterj.nl/2021/09/symfony-6-native-typing#when-upgrading-to-symfony-54) script to automatically add types. Due to the required return types I had to drop PHP 7.4 support.

PHP 7.4 has been removed from the tests and PHP 8.1 has been added instead. Symfony 6.0 has been included as well.

Tested with commit `cd882f0551183d9fb5e75ed56ed4a017181f1572` of the Routing library.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,28 +15,28 @@
         }
     ],
     "require": {
-        "php": "^7.4 || ^8.0",
+        "php": "^8.0",
         "symfony-cmf/routing": "^2.3.0",
-        "symfony/framework-bundle": "^4.4 || ^5.0"
+        "symfony/framework-bundle": "^4.4 || ^5.0 || ^6.0"
     },
     "require-dev": {
         "jackalope/jackalope-doctrine-dbal": "^1.3",
         "doctrine/phpcr-odm": "^1.4|^2.0",
-        "symfony/phpunit-bridge": "^5.0",
+        "symfony/phpunit-bridge": "^5.0 || ^6.0",
         "matthiasnoback/symfony-dependency-injection-test": "^4.1.0",
         "matthiasnoback/symfony-config-test": "^4.1.0",
         "doctrine/orm": "^2.5",
         "symfony-cmf/testing": "^4.0",
         "doctrine/data-fixtures": "^1.0.0",
-        "symfony/form": "^4.4 || ^5.0",
-        "symfony/translation": "^4.4 || ^5.0",
-        "symfony/validator": "^4.4 || ^5.0",
-        "symfony/security-bundle": "^4.4 || ^5.0",
+        "symfony/form": "^4.4 || ^5.0 || ^6.0",
+        "symfony/translation": "^4.4 || ^5.0 || ^6.0",
+        "symfony/validator": "^4.4 || ^5.0 || ^6.0",
+        "symfony/security-bundle": "^4.4 || ^5.0 || ^6.0",
         "doctrine/doctrine-bundle": "^2.0",
-        "symfony/twig-bundle": "^4.4 || ^5.0",
+        "symfony/twig-bundle": "^4.4 || ^5.0 || ^6.0",
         "symfony/monolog-bundle": "^3.5",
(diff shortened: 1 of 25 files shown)


composer.json
```diff
@@ -15,28 +15,28 @@
...
+        "php": "^8.0",
         "symfony-cmf/routing": "^2.3.0",
-        "symfony/framework-bundle": "^4.4 || ^5.0"
+        "symfony/framework-bundle": "^4.4 || ^5.0 || ^6.0"
     },
     "require-dev": {
         "jackalope/jackalope-doctrine-dbal": "^1.3",
         "doctrine/phpcr-odm": "^1.4|^2.0",
-        "symfony/phpunit-bridge": "^5.0",
+        "symfony/phpunit-bridge": "^5.0 || ^6.0",
```
dbu, line 25:
```suggestion
        "symfony/phpunit-bridge": "^6.0",
```

the bridge is only to help with phpunit, no need to support multiple versions.

src/Doctrine/Orm/Route.php
```diff
@@ -35,7 +35,7 @@ class Route extends RouteModel
      *
      * @return self
      */
-    public function setName($name)
+    public function setName($name): \Symfony\Cmf\Bundle\RoutingBundle\Doctrine\Orm\Route
```
dbu, line 38: i'd prefer to have such classes with a `use` statement.

src/Doctrine/Phpcr/PrefixInterface.php
```diff
@@ -22,11 +22,11 @@ interface PrefixInterface
     /**
      * @return string the full path of this document in the repository
      */
-    public function getId();
+    public function getId(): string;
```
dbu, line 25: changing the signature of interface methods is a BC break.

i agree that we should eventually do that, but if we want to quickly release symfony 6 support, we should do that as a minor version without BC breaks.

dbu (commented):
thanks for looking into this.

my preferred path would be to have symfony 6 support as version 2.6 without BC breaks. that way it would also be more visible which changes are due to symfony 6, and which are adding property type, parameter type and return type declarations to the routing bundle things. those changes are BC breaks and will require a new major version. can you split that up please?

trsteel88:
@dbu, any work in progress to make this compatible with Symfony 6?

dbu:
> @dbu, any work in progress to make this compatible with Symfony 6?

i got no reaction on my comments here. i am not aware of any other efforts. glad if you want to give it a go (i'd prefer to not add more return type than absolutely necessary for now, the maximal return type / property type things we should do as a new major version as it is potential BC breaks)

trsteel88:
> > @dbu, any work in progress to make this compatible with Symfony 6?
> 
> i got no reaction on my comments here. i am not aware of any other efforts. glad if you want to give it a go (i'd prefer to not add more return type than absolutely necessary for now, the maximal return type / property type things we should do as a new major version as it is potential BC breaks)

I don't think we can release a minor version with Symfony 6.0 because of all the type hints. We would need to release a major version with the changes that @firstred has made?

dbu:
i don't think we need those type declarations for symfony 6. only in places where we implement an interface method or overwrite a parent method. i am not sure if any of that is a BC break. when its only the DI classes they are not expected to be extended so should be ok. but would need to have a look.

trsteel88:
> i don't think we need those type declarations for symfony 6. only in places where we implement an interface method or overwrite a parent method. i am not sure if any of that is a BC break. when its only the DI classes they are not expected to be extended so should be ok. but would need to have a look.

Was pretty straightward with 5 changes - https://github.com/symfony-cmf/routing-bundle/pull/472

I didn't modify any tests.

PR #472, opened 2022-01-20
Add return types for sf6

trsteel88:
Added required return types for symfony 6.

getContentId() was also updated to return null (instead of void). The phpdoc indicated this should be string|null

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "matthiasnoback/symfony-dependency-injection-test": "^4.1.0",
         "matthiasnoback/symfony-config-test": "^4.1.0",
         "doctrine/orm": "^2.5",
-        "symfony-cmf/testing": "^4.1",
+        "symfony-cmf/testing": "dev-master as 4.2.0",
         "doctrine/data-fixtures": "^1.0.0",
         "symfony/form": "^6.0",
         "symfony/translation": "^6.0",
--- a/composer.json
+++ b/composer.json
@@ -15,8 +15,8 @@
         }
     ],
     "require": {
-        "php": ">=8.0.2",
-        "symfony-cmf/routing": "^2.3.2",
+        "php": "^8.0",
+        "symfony-cmf/routing": "dev-master as 3.0.0",
         "symfony/framework-bundle": "^6.0"
     },
     "require-dev": {
--- a/composer.json
+++ b/composer.json
@@ -15,28 +15,28 @@
         }
     ],
     "require": {
-        "php": "^7.4 || ^8.0",
+        "php": ">=8.0.2",
         "symfony-cmf/routing": "^2.3.2",
-        "symfony/framework-bundle": "^4.4 || ^5.0"
+        "symfony/framework-bundle": "^6.0"
     },
     "require-dev": {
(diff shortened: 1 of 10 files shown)


dbu:
thanks a lot! this makes me realize that there is no way around the BC break and doing a new major version for symfony 6. our Route model extends the symfony route object, and that model is supposed to be an extension point to implement your own stored routes...

could you make this a major version bump (adjust the branch-alias in composer.json and update the CHANGELOG accordingly?). and drop PHP 7.x support, have all builds use PHP 8.0 or 8.1 so that static works. if we have to drop older symfony versions for that to work, lets drop those as well.

i would still prefer to have just the steps necesary to work with symfony 6 in this PR, and then rebase #471 to have that one add strict typing and return type declarations to our own code, for a clearer git history.

trsteel88:
@dbu, update composer.json and the test workflow. However, I think a new version of the routing component will need to be released first (as it does not support symfony 6)

composer.json
```diff
@@ -15,28 +15,28 @@
         }
     ],
     "require": {
-        "php": "^7.4 || ^8.0",
+        "php": ">=8.0.2",
```
dbu, line 18: i know symfony itself does this, but i strongly oppose. we have no clue what php 9 will be like, and chances are reasonable that there will be BC breaks. if we don't limit to the major version, this code might be happily installed by composer in a php 9 application and not work as expected.

composer.json
```diff
@@ -15,28 +15,28 @@
...
         "symfony-cmf/routing": "^2.3.2",
```
dbu, line 19:
```suggestion
        "symfony-cmf/routing": "dev-master as 3.0.0",
```

https://github.com/symfony-cmf/Routing/compare/2.3.4...master - we started on this but did not release yet.

probably need to go over the code to see if there are other cleanups - but would be good to see the build with the bundle as well, to prepare the release of both together

dbu (changes requested):
thanks

composer.json
trsteel88: Ah of course! That's been updated to ^8.0 (the Symfony Framework requirements will ensure this is >=8.0.2)

composer.json
trsteel88: Thanks, that's been updated

trsteel88:
@dbu, it looks like just the prefer lowest test is failing. I think because some Symfony 5.4 packages are being installed. Should this be scrapped from the matrix? Or, how do enforce those packages to use >=6.0?

dbu:
the lowest version build proved its worth here. it is meant to prevent insallation of this library with incompatible components. i tried some things and found that a too old doctrine/common does not handle the `: static` return type declaration correctly on proxies. so i had to adjust the constraint and now the lowest build is fine too.

dbu:
merged in #473, thanks a lot!

trsteel88:
No worries @dbu. In case you missed it, the testing bundle also needs a bump. I used dev-master as 4.2.0.

dbu:
i am aware. currently trying to do the static typing on the routing library so that we then can release it, and then will see about releasing testing and then finally the bundle.

PR #479, opened 2022-04-08
Fix #478: Override getRouteKey on the Route Entity (ORM) to return th…

### LINKED ISSUE #478: ORM: _route attribute of request set to ID rather than to name
### Environment
PHP 8.1

#### Symfony packages

```
symfony/* 6.0.*
```

#### Symfony CMF packages

```
symfony-cmf/routing        3.0.0 3.0.0 Extends the Symfony routing component for dynamic routes and chaining several routers
symfony-cmf/routing-bundle 3.0.0 3.0.0 Symfony RoutingBundle
```

## Subject

When ORM implementation of the Route entity is used, the "_route" attribute of the Request object is set to the numeric ID of the entity (e.g. "5"). What makes impossible to use this attribute along with the URL generator (cannot generate the URL from the ID, it must be the name).

After investigation I spotted it's due to the fact, that `Symfony\Cmf\Bundle\RoutingBundle\Model::getId()` always returns `string`. Before there was no casting to string. 
The above with the if statement in `Symfony\Cmf\Component\Routing\NestedMatcher::getAttributes` (`is_string($route->getRouteKey())`) replaces the route name with it's numeric ID. 

## Steps to reproduce

Read the "_route" attribute of Request object when ORM used.

## Expected results

"_route" attribute should contain the route name

## Actual results

"_route" attribute contains the numeric ID of the Route

Proposed solution is to override `getRouteKey` method of `Symfony\Cmf\Bundle\RoutingBundle\Doctrine\Orm\Route` to return the name rather than ID.

dbojdo:
…route name rather than the numeric ID.

| Q             | A
| ------------- | ---
| Branch?       | "master" for new features / the branch of the current release for fixes
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? |no
| Fixed tickets | #478
| License       | MIT
| Doc PR        | reference to the documentation PR, if any

### DIFF
--- a/src/Doctrine/Orm/Route.php
+++ b/src/Doctrine/Orm/Route.php
@@ -57,4 +57,12 @@ class Route extends RouteModel
     {
         return $this->position;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getRouteKey(): string
+    {
+        return $this->getName();
+    }
 }


dbu:
thanks!

PR #500, opened 2025-12-03
allow installation with symfony 8

dbu:
| Q             | A
| ------------- | ---
| Branch?       | 3.x
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

test with symfony 8

* [x] blocked by https://github.com/doctrine/phpcr-odm/pull/879
* [x]  blocked by https://github.com/doctrine/DoctrinePHPCRBundle/pull/410 (sf 8 support in phpcr-bundle)
* [x] symfony-cmf/testing needs to allow symfony 8
* [x] we probably need to rewrite all service configuration from xml to php (use [this tool](https://github.com/GromNaN/symfony-config-xml-to-php) to automate that mostly)


Djoys:
Hello, just for your information, the current (3.1.2) version of CMF routing bundle, and probably all version not compatible with symfony 8, are also not compatible with symfony 7.4 as Symfony has already removed XmlFileLoader in this version.

The requires should be modified to exclude the 7.4 version of symfony (it is currently set to symfony/framework-bundle: ^6.4 || ^7.0)

Thanks for this very useful bundle !

alexander-schranz:
@Djoys thats not the case XML loaders still work on 7.4, else it would be a bc break which symfony does not in minor versions. PS: framework-bundle is just the glue part, make sure to use symfony/dependency-injection and symfony/config of 7.4.

dbu:
what might be happening is that the symfony/dependency-injection component gets installed at version 8. i'd recommend to use symfony flex and configure that you want symfony 7.4. the flex component makes sure that every symfony component - including the indirect dependencies - is installed at the same version to prevent such hickups.

also, i picked this up again to hopefully be able to merge it soon.

composer.json
```diff
@@ -17,32 +17,31 @@
     "require": {
         "php": "^8.1",
         "symfony-cmf/routing": "^3.0.3",
-        "symfony/framework-bundle": "^6.4 || ^7.0"
+        "symfony/framework-bundle": "^8.0"
```
dbu: adjust before merging, only to test enforced sf 8

composer.json
```diff
@@ -17,32 +17,31 @@
...
     "require-dev": {
-        "doctrine/data-fixtures": "^1.0.0",
-        "doctrine/doctrine-bundle": "^2.8",
-        "doctrine/orm": "^2.9 || ^3.0.1",
-        "doctrine/phpcr-bundle": "^3.0",
-        "doctrine/phpcr-odm": "^2.0",
+        "doctrine/data-fixtures": "^1.2 || ^2.0",
+        "doctrine/doctrine-bundle": "^2.8 || ^3.0",
+        "doctrine/orm": "^2.9 || ^3.0.1 || ^4.0",
+        "doctrine/phpcr-bundle": "dev-sf-versions",
```
dbu: adjust before merging

composer.json
```diff
@@ -17,32 +17,31 @@
...
+        "matthiasnoback/symfony-config-test": "^6.2",
+        "symfony/form": "^6.4 || ^7.0 || ^8.0",
+        "symfony/monolog-bundle": "^3.5 || ^4.0",
+        "symfony/security-bundle": "^6.4 || ^7.0 || ^8.0",
+        "symfony/serializer": "^6.4 || ^7.0 || ^8.0",
+        "symfony/translation": "^6.4 || ^7.0 || ^8.0",
+        "symfony/validator": "^6.4 || ^7.0 || ^8.0",
+        "symfony/var-exporter": "^6.4 || ^7.0 || ^8.0",
+        "symfony/twig-bundle": "^6.4 || ^7.0 || ^8.0",
+        "symfony-cmf/testing": "dev-data-fixtures",
```
dbu: adjust before merging

composer.json
```diff
@@ -17,32 +17,31 @@
...
-        "symfony/framework-bundle": "^6.4 || ^7.0"
+        "symfony/framework-bundle": "^6.4 || ^7.0 || ^8.0"
     },
     "require-dev": {
-        "doctrine/data-fixtures": "^1.0.0",
-        "doctrine/doctrine-bundle": "^2.8",
-        "doctrine/orm": "^2.9 || ^3.0.1",
+        "doctrine/data-fixtures": "^1.2 || ^2.0",
+        "doctrine/doctrine-bundle": "^2.8 || ^3.0",
+        "doctrine/orm": "^2.9 || ^3.0.1 || ^4.0",
```
alexander-schranz: there exist no orm 4 yet


```suggestion
        "doctrine/orm": "^2.9 || ^3.0.1",
```

dbu: oops, right, confused myself while trying to get composer install on php 8.5.

we should not claim to support it, or there might be nasty surprises.

alexander-schranz: maybe you want add requirement to [Doctrine\Persistence](https://github.com/doctrine/persistence) @dbu

dbu:
regarding constraint for doctrine/persistence: why do you think we should have one in this bundle? i think it should be up to orm and phpcr-odm to define what they need.

Djoys:
@dbu, @alexander-schranz,

Many thanks for your answers. You are right, i got confused (and also forgot the basics of Symfony BC policy).

I also see that 3.2.0 has been released and compatible with sf 8. Great job and thanks you !

dbu:
no worries. you asking reminded me that there was still open business here :-)

